PR #3, opened 2023-10-28
Raise error only if status code is 404

LabAsim:
Hello @sinaatalay,

i opened a new pull request with only 1 commit, which incorporates the changes that you mentioned.
The function will only raise a ValueError only if the status code is 404 (Not found).


sinaatalay:
Thank you for finding the bug and fixing it. This is the first public contribution to RenderCV  😃

LabAsim:
It's also my first merged pull request, thank you!

PR #9, opened 2023-11-26
Remove local tinytex

michael-sbarra:
- Removes the local tinytex and adds instructions for including as external dependency. 
- Changes README.md to docs/index.md symlink to eliminate duplication


sinaatalay:
Thank you for the pull request! Even though I agree with what you're doing, I don't want to increase the number of steps in the quick start guide. I think being able to use RenderCV right after `pip install rendercv` makes it very convenient for most users. People might want to test the package it quickly. More steps might change their mind about trying the package.

I will do more research about non-Python dependencies for Python projects.

Or maybe (if we are sure that we don't have any other choice because I am not a fan of this approach as well) the first run can install TinyTeX automatically. Fonts can be downloaded with every run if not downloaded before.

sinaatalay:
But I would happily merge the other parts, except for the TinyTeX deletion.

PR #10, opened 2023-11-27
Adds Mastodon social network

jpgoldberg:
The purpose of this PR is to add Mastodon as a social network, resolving issue #11.

Unlike the other social networks supported, there is no fixed hostname for the URL. Instead the hostname for the URL must be constructed from the full Mastodon address. 

That is, a mastodon address may look like "`a_tooter@social.example`" which should then have the associated url be "`https://social.example/@a_tooter`"

## Why is this PR more complicated than one might expect?

This really seems like it would be a simple fix.

### Mastodon specs and DNS hostnames

Unfortunately there is no formal specification for these things. I am grateful @ThisIsMissEm for [pointing me](https://hachyderm.io/@thisismissem/111467925543315876 ) to the relevant lines of [Mastodon reference implementation](https://github.com/mastodon/mastodon). 

Unfortunately, what I glean from that would allow syntactically invalid hostnames. And so I included stricter validation of hostnames. Perhaps I failed to understand the Pydantic and [dnspython documentation](https://dnspython.readthedocs.io/en/stable/name.html)  packages, but I found no obvious way to validate hostnames properly. So I resorted to a [stack exchange answer](https://stackoverflow.com/a/33214423/1304076).

I wasn't sure where to put my Mastodon address handling methods, so they are static methods in data_model.Connections.

### Getting things to run

I need to tinker with `pyproject.toml` to get even the original to build on my system. I do not believe that I've done any harm, but testers should make sure that I haven't messed up build or packaging process.

I also added tests, and to make it easier for me to run tests tinkered with `tests/test_data_model.py`. Again, I don't think I did anything that should interfere with other testing.


sinaatalay:
Thank you! I will review it thoroughly this weekend.

ThisIsMissEm:
Whilst I cannot review this, @jpgoldberg I think what you'll actually want to do is make a webfinger request:

```
https://host.example/.well-known/webfinger?resource=acct:username@host.example
``` 

this'll either succeed or fail, and gives the URLs to any such profile.

e.g., mine:
```
{
  "subject" : "acct:thisismissem@hachyderm.io",
  "aliases" : [
    "https://hachyderm.io/@thisismissem",
    "https://hachyderm.io/users/thisismissem"
  ],
  "links" : [
    {
      "rel" : "http://webfinger.net/rel/profile-page",
      "type" : "text/html",
      "href" : "https://hachyderm.io/@thisismissem"
    },
    {
      "rel" : "self",
      "type" : "application/activity+json",
      "href" : "https://hachyderm.io/users/thisismissem"
    },
    {
      "rel" : "http://ostatus.org/schema/1.0/subscribe",
      "template" : "https://hachyderm.io/authorize_interaction?uri={uri}"
    },
    {
      "rel" : "http://webfinger.net/rel/avatar",
      "type" : "image/jpeg",
      "href" : "https://media.hachyderm.io/accounts/avatars/109/296/581/037/186/439/original/b55ac3f3a27e171e.jpeg"
    }
  ]
}
```

jpgoldberg:
> Whilst I cannot review this, @jpgoldberg I think what you'll actually want to do is make a webfinger request:
> 
> ```
> https://host.example/.well-known/webfinger?resource=acct:username@host.example
> ```

Oh, that is a very cool thing.  I did not know about it. (As you see, I really know very little about the protocols associated with Mastodon.)

But I do not want to take that approach for two (related) reasons

1. I do not want to make a query to `https://host.example/` before knowing that "`host.example` is syntactically valid.  Attempting to _use_ malformed data to see if it breaks as a way to test for its validity is the road to many security bugs. Instead, we should always check that something is syntactically valid before attempting to use it. Making a network call with potentially malformed data is a use, even if the use is intended as a validity check.

2. I would like to be able to construct the CV without a network connection. The CV should build off-line.

I really am a stickler for checking syntactic validity before any use. And, as you see, I do not pass up the opportunity to rant/lecture about it. Checking strict syntactic validity before any use is a really good habit to get into, and it prevents security issues that might otherwise arise years later.

rendercv/data_model.py
```diff
@@ -1048,13 +1048,99 @@ class Connection(BaseModel):
...
+        # the last label must be not all-numeric
+        if re.match(r"[0-9]+$", labels[-1]):
+            return False
+
+        # labels cannot begin with a hyphen
+        # labels must have at least character
+        # labels may not have more than 63 characters
+        allowed = re.compile(r"(?!-)[a-z0-9-]{1,63}(?<!-)$", re.IGNORECASE)
+        return all(allowed.match(label) for label in labels)
+
```
sinaatalay, line 1082: If we need to verify Mastodon usernames, we should also verify other networks. Is it necessary to validate usernames? Moreover, if we want to do a validation, we should use the [@field_validator](https://docs.pydantic.dev/latest/concepts/validators/#field-validators) decorator of Pydantic. This will help exception handling since we will know it's a validation error.

rendercv/data_model.py
```diff
@@ -1048,13 +1048,99 @@ class Connection(BaseModel):
...
+        uname = m.group("uname")
+        domain = m.group("domain")
+
+        # the domain part of pattern allows some things that are not
+        # valid names. So we run a stricter check
+        if not Connection.is_valid_hostname(domain):
+            raise ValueError("Invalid hostname in mastodon address")
+
+        url = HttpUrl(f"https://{domain}/@{uname}")
+        return url
```
sinaatalay, line 1142: I don't see this function being used anywhere other than the `url` method. Maybe we can move the contents inside the `url` method for clarity?

tests/test_data_model.py
```diff
@@ -858,3 +860,46 @@ def test_read_input_file(self):
...
+        mastodon_name = "user@bad.numeric.tld.123"
+        with self.subTest("All digit TLD"):
+            with self.assertRaises(ValueError):
+                data_model.Connection.MastodonUname2Url(mastodon_name)
+
+        mastodon_name = "a_tooter@example.exchange."
+        expected = HttpUrl("https://example.exchange./@a_tooter")
+        result = data_model.Connection.MastodonUname2Url(mastodon_name)
+        with self.subTest("With FQDN root '.'"):
+            self.assertEqual(result, expected)
```
sinaatalay, line 986: This makes sense, and I will consider creating a new test for each social network. However, instead of using the `MastodonUname2Url` method directly, we can create a Connection object and check its `url` field (since we plan to move `MastodonUname2Url` to inside the `url` method.)

sinaatalay (commented):
Hi, this looks great. I would like to merge it. Before the merge, it would be great if we could change a few small things I mentioned in the comments. Let me know if there is something you disagree with. I might have missed something.

rendercv/data_model.py
jpgoldberg: There are two separate suggestions. The reason that Mastodon needs special validation as opposed to, say, LinkedIn is that we know that know that `www.linkedin.com` ia  syntactically valid hostname. We aren't going to create a malicious PDF by using it.

There may be value in syntactically validating other usernames. But I do think it would be a mistake to actually perform any network calls while running RenderCV. So the validation is only syntactic.

I very much suspect you are correct about `@field_validator`, but the be perfectly honest, I don't fully grok what Pydantic is doing. However, I am happy to learn. So I will work on this.

rendercv/data_model.py
jpgoldberg: We have different opinions about "clarity", but it is your project, so I will follow your suggestion.

tests/test_data_model.py
jpgoldberg: Makes sense. Note that we can use Pydantic's HttpUrl class for the urls created for the other social networks.

Indeed, that offers use a simpler way of validating Mastodon domain names, as Pydantic does (mostly) check that the host part of a URL is syntactically valid.

rendercv/data_model.py
sinaatalay: Actually, (if you haven’t started working on it), we can keep it like this for now because the @field_validator check should be done on the `SocialNetwork` class since the `Connection` class is not used by the users directly. But if we move it to `SocialNetwork`, we will need to do something about other social networks, too. Therefore, we can address this validation problem later. Whatever you prefer.

sinaatalay:
Hello, if you're not working on this PR, I would like to merge it as it is.

PR #13, opened 2024-01-12
Proposal for rendering a cover letter.

achugr:
https://github.com/sinaatalay/rendercv/issues/12#issue-2071626586

Please let me know what you think.
The cover letter template is fairly basic at the moment.

[John_Doe_cover_letter.pdf](https://github.com/sinaatalay/rendercv/files/13910908/John_Doe_cover_letter.pdf)

### DIFF
--- a/rendercv/data_model.py
+++ b/rendercv/data_model.py
@@ -610,8 +610,10 @@ class Design(BaseModel):
     @classmethod
     def check_if_theme_exists(cls, theme: str) -> str:
         """Check if the theme exists in the templates directory."""
-        template_directory = str(files("rendercv").joinpath("templates", theme))
-        if f"{theme}.tex.j2" not in os.listdir(template_directory):
+        cv_template_directory = str(files("rendercv").joinpath( "templates", theme, "cv"))
+        cover_letter_template_directory = str(files("rendercv").joinpath( "templates", theme, "cover_letter"))
+        if f"{theme}.tex.j2" not in os.listdir(cv_template_directory) \
+                and f"{theme}.tex.j2" not in os.listdir(cover_letter_template_directory):
             raise ValueError(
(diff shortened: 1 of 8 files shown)


rendercv/data_model.py
```diff
@@ -1179,6 +1181,65 @@ class SectionWithPublicationEntries(SectionBase):
...
+    application_company_name: Optional[LaTeXString] = Field(
+        default=None,
+        title="Application company name",
+        description="The name of the company where you send your application."
+    )
+    letter: Optional[LaTeXString] = None
+
+    @computed_field
+    @cached_property
+    def connections(self) -> list[Connection]:
```
achugr, line 1222: This is a code duplicate, this can be common in some base model.

rendercv/templates/classic/cv/components
```diff
@@ -0,0 +1 @@
+../components
```
achugr, line 1: I was not sure about relative paths resolution, so just added a symlink for now.

sinaatalay:
Hello, thank you for submitting this pull request. However, at the moment, I do not plan to merge it. I have a lot of plans for the CV side and don't want to make things more complicated in the repository right now. I believe RenderCV should focus exclusively on CVs currently.

That being said, I think this is really good. Would you be willing to create a new package called `RenderCoverLetter` as you suggested? If not, I will keep the pull request open. Perhaps in the future, when the CV side is mature enough, we can start developing the cover letter in this code base.

achugr:
Hey, sure, I agree that it's better to do one thing right and stay focused!
I'll figure out how it would look like as a standalone package and if I could reuse utility code without copying it. Conceptually it's not so good, because cover letter package should not depend on the cv package, then it should be some properly organized shared package used by both, which is going to be an unnecessary complexity for your project.
From this perspective, a fork looks easier for me at the moment.

PR #20, opened 2024-02-26
bjorn/translations: Implement localization using Babel

bjornpijnacker:
This PR implements localization for the few hardcoded strings that rendercv contains. The user is able to select from languages in their YAML by using the top-level `language` field. Currently "en" (English) and "nl" (Dutch) is possible, where "en" is the default if not specified.

The translations are implemented using Babel. The following CLI commands are available to manage them (all in the root directory of the project):
  - `pybabel extract -F rendercv/babel-mapping.ini -o rendercv/locale/messages.pot rendercv`
    This extracts translation strings from all the `.py` files and all template files (tex and md) and places them into `messages.pot`.
  - `pybabel update -i rendercv/locale/messages.pot -d rendercv/locale`
    With the new `messages.pot`, sync the changes to the locale files. Now you can add the translations.
  - `pybabel init -l <lang_name> -i rendercv/locale/messages.pot -d rendercv/locale`
  	This adds a new language

I have added a custom hook to the hatch build to compile the `.po` translation files into `.mo` files which are ready for bundling. I have also updated some of the tests since I have changed the date formatting function. All tests pass currently however they do only consider the English version.


sinaatalay:
Thank you very much for this comprehensive pull request!

Since this is a huge update, I will need some time to consider and review this pull request.

bjornpijnacker:
Thank you! I did just find some issues with translation loading myself after doing a clean build and install. I will go ahead and convert to draft until I have fixed it.

Edit: By [f672f2f](https://github.com/sinaatalay/rendercv/pull/20/commits/f672f2f5b83ee981ce050a6200f6596056780b65) it now includes the translation files correctly!

coolusaHD:
@bjornpijnacker  
Nice PR 👍 

@sinaatalay
If you decide to merge this. Just let me know ✌️ I can provide the translations for German

sinaatalay:
RenderCV should be a multi-language tool, but I am thinking of achieving this without any external dependencies.

I think this can be achieved just with `data_models.py`. We can move all the hard-coded strings there and have dictionaries that store the strings for each language. The data models will return strings with computed fields for the templates. If we could achieve that, we wouldn't need `Babel`. Therefore, I don't plan to merge this pull request for now.

Thank you very much again! Please let me know if you think I am missing something.

And @coolusaHD, thank you for the offer! I will let you know when that day comes.

bjornpijnacker:
Thank you for considering. I chose the Babel approach over a Python dictionary approach since Jinja2 has native support for Babel which makes it easier to write templates without changing Python code. I think the Python approach without Babel is also good however, since the requirements for translations in this project is quite simple.

@sinaatalay Do you want me to work on such an approach? In that case I'll close this PR and create a new one once I have time!

sinaatalay:
@bjornpijnacker  Sure that would be very helpful; thanks!

My plan is not to have any hardcoded strings in templates and to keep all the string generation processes in `data_models.py`.

And for the "Last updated in ..." text, I plan to make it a design option like what we currently have for page numbering (`page_numbering_style`), so we don't have to provide the translations for these two.

bjornpijnacker:
I did not end up having time for this in case anybody is reading this still and was deciding whether or not to contribute!

PR #31, opened 2024-03-04
Changes to allow for the sort of publications in ascending, descendin…

lhegstrom:
Adding in the ability to have the list of publications sort in either ascending or descending order based on the date

### DIFF
--- a/rendercv/renderer.py
+++ b/rendercv/renderer.py
@@ -17,7 +17,7 @@ import shutil
 import sys
 import copy
 from datetime import date as Date
-from typing import Optional, Literal, Any
+from typing import Optional, Literal, Any, List, Optional
 
 import jinja2
 import markdown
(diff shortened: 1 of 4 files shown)


sinaatalay:
Thank you for the pull request!

Even though it might be helpful in some cases, I think it breaks the separation between the content and the design RenderCV tries to achieve. The whole content should stay in the `cv` section and not depend on the theme. With this approach, content is updated based on a design option.

Isn’t it trivial for users to keep them ordered in the YAML? Please let me know if I am missing something.

lhegstrom:
That's a fair point. This does break the separation between design content. The motivation is that when transition over to this for those of us with a lot of pubs, it's kind of a massive pain.

Two additional ideas:

- How do you feel about adding a flag to the render cli function similar to the use local latex?

`rendercv render your_yaml.yaml --sort_publications ascending`

- Alternatively.. there could be another specific function / module to assist with yaml changes such as these that keeps the logic / 

`rendercv format your_yaml.yaml --sort_publications ascending`

I'm thinking the second option would be better from a design standpoint. Thoughts?

PR #53, opened 2024-04-07
fix: remove example summary from template

tcinbis:
Hey @sinaatalay,

thank you for this amazing tool, I am really liking what I have seen so far! Just a minor fix since I was experimenting with the latest code changes from master.
I tried to infer what the correct template should look like from your other changes. I hope it is correct.

Best,
Tom

### DIFF
--- a/rendercv/themes/classic/TextEntry.j2.tex
+++ b/rendercv/themes/classic/TextEntry.j2.tex
@@ -3,5 +3,5 @@
 ((* endif *))
 
 \begin{onecolentry}
-    This is an example resume to showcase the capabilities of the open-source LaTeX CV generator, \href{https://github.com/sinaatalay/rendercv}{RenderCV}. A substantial part of the content is taken from \href{https://www.careercup.com/resume}{here}, where a \textit{clean and tidy CV} pattern is proposed by \textbf{Gayle L. McDowell}.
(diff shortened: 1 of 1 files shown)


tcinbis:
Judging from the tests, not quite... but at least with my demo CV the summary worked fine with version 1.6, but now only includes the sample summary that is currently in the `classic/TextEntry.j2.tex` template.

Seems like the testdata needs to be fixed as well.

sinaatalay:
Thank you for finding this! Your approach was spot on. I am merging it.

The tests failed due to mismatches between the RenderCV's output and reference outputs in the `testdata` folder. Since a template has been updated, indeed, the reference outputs in the `testdata` folder should also be updated.

**A note about tests:**

In some of the tests:

- RenderCV generates an output with a sample input.
- Then, the output is compared with a reference output, which has been manually generated and stored in `testdata`. If the files differ, the tests fail.


When the `testdata` folder needs to be updated, it can be manually regenerated by setting `update_testdata` to `True` in `conftest.py` and running the tests.

Whenever the `testdata` folder is generated, the files should be reviewed manually to ensure everything works as expected. But sometimes, mistakes like this one can go unnoticed. Therefore, any ideas about other testing approaches would be appreciated.

PR #67, opened 2024-05-14
Fixed issues with new lines in markdown 

guapman:
I've faced an issue with the new lines in markdown during rendering of my CV. Found that in case if optional fields aren't provided there were redundant new lines. Also project section missed new line after dates.

### DIFF
--- a/rendercv/themes/markdown/EducationEntry.j2.md
+++ b/rendercv/themes/markdown/EducationEntry.j2.md
@@ -1,9 +1,9 @@
 ## <<entry.institution>>, ((* if entry.degree *))<<entry.degree>> in ((* endif *))<<entry.area>>
 
-((* if entry.date_string *))- <<entry.date_string>> ((* endif *))
-
-((* if entry.location *))- <<entry.location>> ((* endif *))
-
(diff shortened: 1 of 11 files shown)


sinaatalay:
Thank you very much! I will merge this but we should fix the tests first.

Tests failed because of the reason I explained in #53:
      
> The tests failed due to mismatches between the RenderCV's output and reference outputs in the `testdata` folder. Since a template has been updated, indeed, the reference outputs in the `testdata` folder should also be updated.
> 
> **A note about tests:**
> 
> In some of the tests:
> 
> - RenderCV generates an output with a sample input.
> - Then, the output is compared with a reference output, which has been manually generated and stored in `testdata`. If the files differ, the tests fail.
> 
> 
> When the `testdata` folder needs to be updated, it can be manually regenerated by setting `update_testdata` to `True` in `conftest.py` and running the tests.
> 
> _Originally posted by @sinaatalay in https://github.com/sinaatalay/rendercv/issues/53#issuecomment-2041646184_

I could also do this but it would be nice for new developers to go through this process. Let me know if you encounter any problems!

guapman:
Yep, my bad. Forgot about tests. I'll check and fix

PR #87, opened 2024-05-29
formatting correction structure_of_the_yaml_input_file.md

averagejoey2000:
Included line breaks before the subheadings for 'normal entry' and 'oneline entry' to make it more visually clear where one ended and the next began.

Without the clear line between, the rendered markdown showed  `highlights: A list of bullet points. {% elif entry_name == "OneLineEntry" %}` and `date: The date as a custom string or in YYYY-MM-DD, YYYY-MM, or YYYY format. {% elif entry_name == "Normal Entry" %}`


sinaatalay:
Thank you very much!

PR #118, opened 2024-07-08
improve boilerplate content in sample_content.yaml

TheColorDarkBlue:
Simplify and improve boilerplate content


sinaatalay:
Thank you. I will merge it, but the tests fail because the website doesn’t have a valid HTTP URL (`https://` has been removed).

The `https` part is never printed on the resume, but we need it to ensure the link's validity in the PDF. So, we should add it back to the YAML. PDF won’t show the `https` part.

TheColorDarkBlue:
> Thank you. I will merge it, but the tests fail because the website doesn’t have a valid HTTP URL (`https://` has been removed).

What about the trailing `/` at the end of certain URLs? Will those automatically be omitted?

> The `https` part is never printed on the resume, but we need it to ensure the link's validity in the PDF. So, we should add it back to the YAML. PDF won’t show the `https` part.

What about `www.` in URLs? Will those also be omitted? For reference, take a look at each of the URLs here https://imgur.com/7lJTTsN

* `portfolio.com` instead of `https://www.portfolio.com/`
* `github.com/name` instead of `https://www.github.com/name/`
* `github.com/name/project` instead of `https://www.github.com/name/project/`
* `project.com` instead of `https://www.project.com/`
* `blog.com/projectname` instead of `https://www.blog.com/projectname/`

I disagree with adding these back to the YAML.

1991CRX:
> > Thank you. I will merge it, but the tests fail because the website doesn’t have a valid HTTP URL (`https://` has been removed).

What about the / at the end of the URL?

> > The `https` part is never printed on the resume, but we need it to ensure the link's validity in the PDF. So, we should add it back to the YAML. PDF won’t show the `https` part.
> 
> Will [www](http://www). in URLs also be omitted? For reference, take a look at each of the URLs here https://imgur.com/7lJTTsN
> 
> * portfolio.com
> * github.com/name
> * github.com/name/project
> * project.com
> * blog.com/projectname

sinaatalay:
`www` is the name of a subdomain; it's up to people whether to use it or omit it. Some people make their website only available to the `www` subdomain. Therefore, it's not being omitted automatically.

Why do you disagree with adding `https`? There is a valid reason for making the URL a valid HTTP URL: We must make sure each link goes to a valid webpage. The PDF looks the same, so I don't understand the issue.

Also, trailing `/` is omitted automatically as well. These can be easily tested with the `rendercv render` command.

TheColorDarkBlue:
> `www` is the name of a subdomain; it's up to people whether to use it or omit it. Some people make their website only available to the `www` subdomain. Therefore, it's not being omitted automatically.
> 
> Why do you disagree with adding `https`? There is a valid reason for making the URL a valid HTTP URL: We must make sure each link goes to a valid webpage. The PDF looks the same, so I don't understand the issue.
> 
> Also, trailing `/` is omitted automatically as well. These can be easily tested with the `rendercv render` command.

I suggest simply making the website entry Markdown

See commit https://github.com/sinaatalay/rendercv/pull/118/commits/0b9456dca82b4917f01415c77465c0988d08023a

sinaatalay:
Maybe you can open an enhancement issue for that, and we can discuss  it under the issue. I don't understand why you would request that since the PDF automatically has the clean URL. I don't understand what you didn't like in the output.

Issue #164, opened 2024-08-29
Filter on content or sections


akib1689:
### Discussed in https://github.com/sinaatalay/rendercv/discussions/163

<sub>Originally posted by **akib1689** on August 29, 2024</sub>

---

I don't know if I'm missing something or not. It would be great if we could hide/include some of the sections from the YAML file.

**My Use Case:**

I apply for 2-3 roles, and each time I need to manually comment/uncomment the relevant portion of the YAML file. It would be very useful if we had something like a tag in each item.

> [!NOTE] 
> Each section will have some tags. For example, one of my skill rows should be included for DevOps-related jobs but not for developer-related jobs. It would be really helpful to set a match tag from the command line, say "devops," and each node that has the tag "devops" would be included in that particular render of the CV. By default, it would include all the contents.

The implementation can be flexible. But here I'm providing one way to implement:

- Each element in our yaml file will have an optional field named `tag`
- This `tag` will be an array of strings.
- Every child element will have parents tag nested format unless they choose to ignore it. one liner currently has one line string then have 2 things say `content`, `tag`  
- Then in cli will have some command line argument say `match-tag` or `include-tag` to include only those nodes that has the that tag attached to them.

akib1689:
I think we should discuss how do implement this section. Here is my proposal:

- Each of the entry types in render cv should have an optional field named `label`. 
- The sections are pre defined. or We can also make the whole section tag able like this. Then we can include whole section by that label. For my case It would be say publications. For job seeking I don't like to include publications section. I need it only for university application. 
-  Contrary to the second point, we might have a business logic that is if the entry of any of the section becomes 0, Then don't include the section. Either one is fine for me.

```yaml
rendercv_settings:
  cli_options: # might be called render options also
    dont_generate_png: true
    # ... other cli options to implement
  versions:
    # array of different version
    - label1:
      include:
        labels: label1  # parses all the contents and its sub contents labled devops only.
```

sinaatalay:
The `cv` field of the YAML input file should be used only for content-writing purposes. Whatever is in there should be the stuff that is meant to be printed in the PDF. What about something like this?

```yaml
rendercv_settings:
  hidden_sections:
    - Publications
```

akib1689:
Sorry, for not getting back to you, I would like this then:

```yaml
rendercv_settings:
  versions:
    - devops:
      - experience.exp1
      - experience.exp2
    - research:
      - experience.exp2
      - experience.exp3
```

[...]

PR #166, opened 2024-09-01
feat: Added the `rendercv_settings` block

### LINKED ISSUE #146: Add `rendercv_settings` field to the data model
In addition to the `cv`, `design`, and `locale_catalog` fields, we should include a new field called `rendercv_settings`.

`rendercv_settings` could contain all the CLI options (like output folder name, pdf path, etc.) and other settings like discussed in #139 and #144.

akib1689:
As discussed in #146, A new block named `rendercv_settings` is added. 

Usage of this block:

```yaml
rendercv_settings:
  output_folder_name: 'str'
  use_local_latex_command: 'str'
  pdf_path: 'str'
  latex_path: 'str'
  html_path: 'str'
  png_path: 'str'
  markdown_path: 'str'
  no_html: false
  no_markdown: false
  no_png: false
```

Each and every property in this block is direct equivalent implementation of the cli option.

fixes: #146


rendercv/cli/utilities.py
```diff
@@ -260,3 +260,35 @@ def parse_render_command_override_arguments(
...
 
     return key_and_values
+
+
+def build_rendercv_settings(
+    dictionary: dict,
+    command_line_arguments: dict[str, str],
+    command_line_arguments_default_values: dict[str, str],
+) -> dict[str, str]:
+    """Build the RenderCV settings dictionary by combining the dictionary and the command line arguments.
```
sinaatalay: Please make sure the maximum width of anything in the source code is below 88 lines. I won't add the same comment for the other ones. You can use a vertical ruler in VS Code.

rendercv/cli/commands.py
```diff
@@ -164,17 +184,8 @@ def cli_command_render(
...
-        number_of_steps = number_of_steps - 1
-    if dont_generate_markdown:
-        # if the Markdown file is not generated, then the HTML file is not generated
-        number_of_steps = number_of_steps - 2
-    else:
-        if dont_generate_html:
-            number_of_steps = number_of_steps - 1
-
-    with printer.LiveProgressReporter(number_of_steps) as progress:
+    initial_steps = 1
```
sinaatalay: Apparently, updating render settings after the YAML file is read complicates some logic. I would try to avoid complicating the logic as much as possible. Here is my proposal:

Before starting the steps, do these first:

-   Figure out which settings are given from CLI, i.e., which are different from the default values.
-   Read the YAML file as a dict, and open the settings as a dict. We don't need to validate yet.
-   Add/overwrite all the given CLI settings to the settings dict, which is read from the YAML file.
-   Then, compute the steps using the old logic and start the steps (start directly from validation).

rendercv/cli/printer.py
```diff
@@ -84,6 +84,22 @@ def start_a_step(self, step_name: str):
             f"{self.current_step_name} has started."
         )
 
+    def update_total_steps(self, total_steps: int):
```
sinaatalay: See https://github.com/sinaatalay/rendercv/pull/166#discussion_r1740264333. We probably don't need this.

rendercv/cli/utilities.py
```diff
@@ -260,3 +260,35 @@ def parse_render_command_override_arguments(
         key_and_values[key] = value
 
     return key_and_values
+
+
+def build_rendercv_settings(
```
sinaatalay: We may need to udpate this according to https://github.com/sinaatalay/rendercv/pull/166#discussion_r1740264333.

Also, I would suggest naming this: `update_render_settings`.

And we don't need to store default values in another dict. See `inspect` module of Python. It can read functions and return the default values of the arguments. If we don't use this, every time we change a default value, we will be in trouble.

rendercv/cli/commands.py
```diff
@@ -141,18 +141,38 @@ def cli_command_render(
...
+        "markdown_path": utilities.string_to_file_path(markdown_path),
+        "html_path": utilities.string_to_file_path(html_path),
+        "png_path": utilities.string_to_file_path(png_path),
+        "no_markdown": dont_generate_markdown,
+        "no_html": dont_generate_html,
+        "no_png": dont_generate_png,
+    }
+
+    # Create the default values for the cli_args:
+    cli_args_default = {
```
sinaatalay: See https://github.com/sinaatalay/rendercv/pull/166#discussion_r1740265891

rendercv/data/models/rendercv_data_model.py
```diff
@@ -36,6 +37,11 @@ class RenderCVDataModel(RenderCVBaseModelWithoutExtraKeys):
         ),
         validate_default=True,
     )
+    rendercv_settings: RenderCVSettings = pydantic.Field(
```
sinaatalay: Please make it `Optional` and assign `None` by default so that it doesn't show up in the sample YAMLs (like `locale_catalog`).

rendercv/data/models/rendercv_settings.py
```diff
@@ -0,0 +1,134 @@
...
+
+import pydantic
+
+
+class RenderCVSettings(pydantic.BaseModel):
+    """This class is the data model of the rendercv settings. The values of each field
+    updates the `rendercv_settings` dictionary.
+    """
+
+    model_config = pydantic.ConfigDict(
```
sinaatalay: Never use `model_config` and always inherit from `base.py`. See the other classes.

rendercv/data/models/rendercv_settings.py
```diff
@@ -0,0 +1,134 @@
...
+The `rendercv.models.rendercv_settings` module contains the data model of the
+`rendercv_settings` field of the input file.
+"""
+
+from typing import Optional
+
+import pydantic
+
+
+class RenderCVSettings(pydantic.BaseModel):
```
sinaatalay: Please update `RenderCVSettings` so that the YAML file looks like this:

```yaml
rendercv_settings:
  render_settings: # or maybe just "render:" ?
    ... # all available render command CLI options
```

rendercv/renderer/renderer.py
```diff
@@ -12,7 +12,6 @@
 from typing import Optional
 
 import fitz
```
sinaatalay, line 14: Make sure to use

```
hatch run default:sort-imports
```

and

```
hatch run default:format
```

See the developer guide for details.

sinaatalay:
Thanks a lot for your great work! It looks really good. I've included some comments.

I'm quite tied up these days, so I might not be able to respond immediately, but I'll get back to you as soon as I can.

sinaatalay:
When everything is all set, we need to write tests for every function created. Overriding the settings from YAML with CLI should be tested, too.

sinaatalay (changes requested):
When everything is all set, we need to write tests for every function created. Overriding the settings from YAML with CLI should be tested, too.

rendercv/renderer/renderer.py
akib1689: When I run:

```bash
hatch run default:sort-imports
```

It generates back to this format. May be there are some problem with my configuration? I'm not being able to figure this out.

akib1689:
I'll Use the comments as check-boxes and resolve them as I implement them. Is it okay? Once I resolve all the comments I'll use the ready for review.

rendercv/cli/printer.py
akib1689: Deleted the function

rendercv/cli/commands.py
akib1689: Moved the reading logic to the top.

Now our 1st step only verifies the dictionary. Not reads them.

rendercv/data/models/rendercv_data_model.py
akib1689: Made the field optional like locale catalog.

But I have some question:

```python
@pydantic.field_validator("locale_catalog")
    @classmethod
    def initialize_locale_catalog(cls, locale_catalog: LocaleCatalog) -> LocaleCatalog:
        """Even if the locale catalog is not provided, initialize it with the default
        values."""
        if locale_catalog is None:
            LocaleCatalog()

        return locale_catalog

    @pydantic.field_validator("rendercv_settings")
    @classmethod
    def initialize_rendercv_settings(
            cls, 
            rendercv_settings: RenderCVSettings
        ) -> RenderCVSettings:
        """Even if the rendercv settings are not provided, initialize them with 
        the default values."""
        if rendercv_settings is None:
            RenderCVSettings()

        return rendercv_settings
```

Here the `LocaleCatalog()` and `RenderCVSettings()` statements shows the instruction is unreachable. Is there anything wrong I configured differently?

rendercv/cli/utilities.py
akib1689: Should I change the variable name of the function to keep consistency with the data model?

If that is okay. then I'll do this:

- `dont_generate_markdown` variable will be called `no_markdown`
- `dont_generate_html` variable will be called `no_html`
- `dont_generate_png` variable will be called `no_png`

These changes will be present in the code only. or should I change the rendercv settings blocks definition?

rendercv/data/models/rendercv_settings.py
akib1689: Done.

rendercv/data/models/rendercv_settings.py
akib1689: This might be a good Idea. Later we can add `filter` or `version` to reflect #164

akib1689: This is done. 

The render cli settings can be found in the following format:

```yaml
rendercv_settings:
  render:
    output_folder_name:
    # ... other options
```

rendercv/cli/commands.py
akib1689: resolved

rendercv/cli/utilities.py
akib1689: Done this issue

akib1689:
@sinaatalay  Can you please review the changes now? I think most of the issue is resolved. and after your answer on the unresolved comments. I'll update and make the pr final.

rendercv/cli/utilities.py
sinaatalay: I prefer the old names. We can update the YAML file inputs with the old names.

rendercv/cli/commands.py
```diff
@@ -141,57 +142,90 @@ def cli_command_render(
...
+        "markdown_path": utilities.string_to_file_path(markdown_path),
+        "html_path": utilities.string_to_file_path(html_path),
+        "png_path": utilities.string_to_file_path(png_path),
+        "no_markdown": no_markdown,
+        "no_html": no_html,
+        "no_png": no_png,
+    }
+
+    # Use inspect to get the default values of the arguments:
+    sig = inspect.signature(cli_command_render)
```
sinaatalay: The `inspect` package should be moved inside the function `update_render_settings`.

rendercv/cli/commands.py
```diff
@@ -141,57 +142,90 @@ def cli_command_render(
...
+    if extra_data_model_override_argumets:
+        key_and_values = dict()
+        key_and_values = utilities.parse_render_command_override_arguments(
+            extra_data_model_override_argumets
+        )
+        data_as_a_dict = utilities.set_or_update_values(data_as_a_dict, key_and_values)
+
+    # update the data of the rendercv settings:
+    render_cv_settings = data_as_a_dict.get("rendercv_settings", dict())
+    if render_cv_settings is None:
```
sinaatalay: Lines 196 and 197 are not required since `get` method will return `dict()` if `rendercv_settings` is not found in the dictionary.

rendercv/cli/commands.py
```diff
@@ -141,57 +142,90 @@ def cli_command_render(
...
+        render_cv_settings = dict()
+
+    # get the render options:
+    render_options = render_cv_settings.get("render", dict())
```
sinaatalay: I would use `data_as_a_dict.get("rendercv_settings", dict()).get("render", dict())` directly. Moreover, you can move this inside `update_render_settings`.

rendercv/cli/commands.py
```diff
@@ -141,57 +142,90 @@ def cli_command_render(
...
+    render_options = utilities.update_render_settings(
```
sinaatalay: `update_render_settings` should get two inputs: `data_as_a_dict` and `cli_args`. Then, it should return the new `data_as_a_dict`. In here, there shouldn't be any logic related to this (`inspect` package, checking if user gave settings in the YAML, etc.)

rendercv/data/models/render.py
```diff
@@ -0,0 +1,131 @@
...
+"""
+
+from typing import Optional
+
+import pydantic
+
+from .base import RenderCVBaseModelWithExtraKeys
+
+
+class RenderOptions(RenderCVBaseModelWithExtraKeys):
```
sinaatalay: You can move all this to `rendercv_settings`. We don't need a seperate file for now.

rendercv/data/models/rendercv_data_model.py
```diff
@@ -36,6 +37,11 @@ class RenderCVDataModel(RenderCVBaseModelWithoutExtraKeys):
         ),
         validate_default=True,
     )
+    rendercv_settings: Optional[RenderCVSettings] = pydantic.Field(
+        default=None,
+        title="RenderCV Settings",
+        description="The settings of the RenderCV.",
```
sinaatalay, line 44: `validate_default` should be set to `True`.

sinaatalay (changes requested):
Thank you! I have made a few more comments; then it will be ready. But we need tests. If you don't want to implement tests, I can do that too. If you want to work on it, let me know.

rendercv/data/models/rendercv_data_model.py
sinaatalay: Interesting. I will have a look.

rendercv/cli/commands.py
akib1689: This is specifically for one single use case:

```yaml
rendercv_settings:
  # .. empty block no 'render' key provided
```

rendercv/data/models/render.py
akib1689: I would argue here. I would immediately start working on #164 and this implementation would be more helpful for that task

rendercv/data/models/rendercv_data_model.py
akib1689: done

rendercv/cli/commands.py
akib1689: This creates a circular dependency. 

```python
import utilitis # In commands
```

```python
import commands # In untilities
```

I am using import statement in the function

rendercv/cli/commands.py
akib1689: Done

akib1689:
> Thank you! I have made a few more comments; then it will be ready. But we need tests. If you don't want to implement tests, I can do that too. If you want to work on it, let me know.

It would be very helpful. If you can implement testing as I'm not familiar with testing tools used in this project. But I'll surely can do documentation. All the change is implemented now. Only the documentation is remaining.

rendercv/data/models/render.py
sinaatalay: There are too many files now. We might consider folders inside the `models` folder (but maybe in another PR).

akib1689: Okay.

rendercv/cli/commands.py
sinaatalay: Can you use dependency injection instead? i.e., the function should take the `cli_command_render` function itself as input. Using imports inside the function might be fine, too, but dependency injection is more explicit and self-contained.

akib1689: I suggest doing this in a later (chore) titled PR. I need to read more on how to do this in python.

sinaatalay: Sure, let me know when you are done with the PR.

sinaatalay:
Let me know when you have finished the pull request following our last discussion.

akib1689:
@sinaatalay Please review the documentation. and overall code structure. I think the pull request is ready for merging now.

sinaatalay (approved):
Thank you for your work! This is the most comprehensive contribution so far.

PR #170, opened 2024-09-07
cli: implement `--watch` cli option

Criyl:
Love the project!

Started working with it and found myself wanting to generate content on save.


Criyl:
resolved conflict with pyproject.toml

sinaatalay:
Thanks a lot. I'll review it as soon as I can.

rendercv/cli/watcher.py
```diff
@@ -0,0 +1,76 @@
...
+        self.on_modified(FileModifiedEvent(src_path=str(self.file_path)))
+
+    def on_modified(self, event: FileModifiedEvent) -> None:
+        if event.src_path != str(self.file_path):
+            # Ignore any events that aren't our file.
+            return
+
+        file_hash = sha256(open(event.src_path).read().encode("utf-8")).hexdigest()
+
+        if file_hash == self.previous_hash:
```
sinaatalay: If `on_modified` is only called when a file has been modified, why are we still checking if it has been changed?

rendercv/cli/commands.py
```diff
@@ -136,6 +144,32 @@ def cli_command_render(
...
+                html_path=html_path,
+                png_path=png_path,
+                dont_generate_markdown=dont_generate_markdown,
+                dont_generate_html=dont_generate_html,
+                dont_generate_png=dont_generate_png,
+                watch=False,
+                extra_data_model_override_argumets=extra_data_model_override_argumets,
+            )
+
+        file_path = utilities.string_to_file_path(input_file_name)
```
sinaatalay: In v1.14 we removed this function from `utilities`. It should be `input_file_path: pathlib.Path = pathlib.Path(input_file_name).absolute()` like line 175.

rendercv/cli/watcher.py
```diff
@@ -0,0 +1,76 @@
...
+from hashlib import sha256
+from typing import Callable
+
+from watchdog.events import FileModifiedEvent, FileSystemEventHandler
+from watchdog.observers import Observer
+from typer import Exit
+
+
+class ModifiedCVEventHandler(FileSystemEventHandler):
+    """This class handles the file changes and triggers a specified `callback` ignoring duplicate changes.
```
sinaatalay: Please make sure the maximum width of anything in the source code is below 88 lines. You can use a vertical ruler in VS Code.

rendercv/cli/watcher.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.previous_hash = file_hash
+
+        try:
+            self.callback()
+        except Exit:
+            ...  # Suppress typer Exit so we can continue watching even if we see errors.
+
+
+def watch_file(file_path: pathlib.Path, callback: Callable[..., None]):
```
sinaatalay: Sorry, this is just my personal preference, but I use very explicit naming in RenderCV. I would use these, but it's up to you:

- Instead of `watch_file`, `run_a_function_if_a_file_changes`.
- Instead of argument `callback`, `function`.

rendercv/cli/commands.py
```diff
@@ -136,6 +144,32 @@ def cli_command_render(
...
+        if file_path is None:
```
sinaatalay: This is already handled. We don't need 168 and 169.

rendercv/cli/commands.py
```diff
@@ -123,6 +123,14 @@ def cli_command_render(
             help="Don't generate the PNG file.",
         ),
     ] = False,
+    watch: Annotated[
+        bool,
+        typer.Option(
+            "--watch",
+            "-w",
+            help="Automatically generate files on change.",
```
sinaatalay: I would write: "Automatically re-run RenderCV when the input file is updated."

rendercv/cli/commands.py
```diff
@@ -123,6 +123,14 @@ def cli_command_render(
             help="Don't generate the PNG file.",
         ),
     ] = False,
+    watch: Annotated[
```
sinaatalay, line 125: We need to append this option to the `cli_render_arguments` below. Also, we need to add a new field to the `RenderCommandSettings` class in the `data.models.rendercv_settings`.

sinaatalay (changes requested):
Thank you for your work! This is an excellent idea. I will merge this, but I have left some comments. Also, we need to write tests for this. Would you be interested in doing that? I can answer any questions you have.

rendercv/cli/watcher.py
Criyl: The on_modified event can trigger regardless of the contents. My use case was that these events occur on each save of a file in VS Code but the same problem occurs in other editors like Vim.

Criyl:
I addressed your comments and added tests for the watcher specific code. Still working on the test for the argument itself.

Criyl:
Sorry for not coming back to this for a hot minute. 

I want to smoke test the watch functionality on a Mac since the tests seem to be failing on MacOS. If anyone has a mac and can manually test the functionality I would appreciate it.

sinaatalay:
I am using Mac. I will see what's going on in my first availability.

Criyl:
> I am using Mac. I will see what's going on in my first availability.

Looks like Mac needs to watch the directory not the file, just like with windows.

I added a better test that actually ensures the render pass is done according to the file edits. So the PR should be ready to review at your earliest convenience.

Love to see the green.

rendercv/cli/commands.py
```diff
@@ -136,6 +144,30 @@ def cli_command_render(
...
+                dont_generate_markdown=dont_generate_markdown,
+                dont_generate_html=dont_generate_html,
+                dont_generate_png=dont_generate_png,
+                watch=False,
+                extra_data_model_override_argumets=extra_data_model_override_argumets,
+            )
+
+        input_file_path: pathlib.Path = pathlib.Path(input_file_name).absolute()
+        watcher.run_a_function_if_a_file_changes(input_file_path, rerun_command)
+        return
```
sinaatalay: We should move this block after 205 so that users can use the `watch` from the YAML file, not necessarily from the CLI. It should look for `render_command_settings_dict["watch"]`, not `watch` directly.

rendercv/cli/watcher.py
sinaatalay: Is there documentation somewhere that lists the events that trigger `on_modified` exactly? It makes sense to me to be triggered when a file is saved; it implies it has changed, right?

rendercv/cli/watcher.py
```diff
@@ -0,0 +1,84 @@
...
+        self.previous_hash = file_hash
+
+        try:
+            self.function()
+        except Exit:
+            ...  # Suppress Exit so we can continue watching.
+
+
+def run_a_function_if_a_file_changes(
+    file_path: pathlib.Path, function: Callable[..., None]
```
sinaatalay: I don't understand the type `Callable[..., None]`. Why isn't just `Callable` enough?

rendercv/cli/watcher.py
```diff
@@ -0,0 +1,84 @@
...
+):
+    """Watch file located at `file_path` and trigger callback on file modification.
+
+    Args:
+        file_path (pathlib.Path): The path of the file to watch for.
+        function (Callable[..., None]): The function to be called on file modification.
+        *FUNCTION MUST BE NON-BLOCKING*
```
sinaatalay: It would be very nice if we explained `FUNCTION MUST BE NON-BLOCKING1` a little more.

tests/test_cli.py
```diff
@@ -922,3 +959,75 @@ def test_render_command_overriding_input_file_settings(
...
+    mocker = WriteEventMocker(new_input_file_path, writes)
+    
+    with patch('time.sleep', side_effect=mocker.side_effect):
+        try:
+            result = runner.invoke(
+                    cli.app,
+                    [
+                        "render",
+                        str(new_input_file_path.relative_to(tmp_path)),
+                        f"-w"
```
sinaatalay: Ruff linter issue here.

sinaatalay (changes requested):
I made a few important comments. Then, it will be ready to be merged. Also, I didn't understand how tests work. Which function mocks which one, what does `assert_condition` do, what is `side_effect`, etc.? It would be nice if they were explained or named a bit differently. Thank you!

I can make these changes by myself too btw, whichever you prefer.

rendercv/cli/watcher.py
Criyl: The [..., None] just specifies that the function can take anything and return nothing. But we dont expect any parameters at this time so Callable or Callable[None, None] would be more appropriate.

rendercv/cli/watcher.py
Criyl: The documentation for the [on_modified](https://pythonhosted.org/watchdog/api.html#watchdog.events.FileSystemEventHandler.on_modified) event says.

> Called when a file or directory is modified.

If you want I can make a decorator to split the logic out.

```py
@cached__from_last_file
def on_modified(self, event: FileModifiedEvent) -> None:
    ...
 ```
 
 Might actually look like below so we can reference the desired file_path
```py
def on_modified(self, event: FileModifiedEvent) -> None:
    with cached_from_last_file_event(self.file_path):
        ...
 ```

rendercv/cli/watcher.py
Criyl: After further consideration I dont think this disclaimer is a concern. Any function can go in here and behave how the user would expect it to.

Criyl:
> I made a few important comments. Then, it will be ready to be merged. Also, I didn't understand how tests work. Which function mocks which one, what does `assert_condition` do, what is `side_effect`, etc.? It would be nice if they were explained or named a bit differently. Thank you!
> 
> I can make these changes by myself too btw, whichever you prefer.

I clarified some things in function names and docstrings.

sinaatalay:
Hello @Criyl, can I push some commits before merging this?

Criyl:
> Hello @Criyl, can I push some commits before merging this?

Ofcourse!

sinaatalay:
Thanks a lot again. I've made some changes in [`4b1deb2`](https://github.com/rendercv/rendercv/pull/170/commits/4b1deb223c26e1c953600677d3d697da4f3c392c). If you think I missed something, please let me know.

Issue #74, opened 2024-05-18
Answer an FAQ: Can I add a profile picture?


sinaatalay:
I have added a FAQ section to the user guide. Your assistance with this is greatly appreciated.

edavalosanaya:
This feature would be great!

kito323:
Using `moderncv` theme it is actually quite easy. After first generation you have the `moderncv/Preamble.j2.tex` file. In this you can follow any moderncv LaTeX code basically and they have a `\photo` command:
> \photo A command for a photo. Takes the image file name as a required argument. Takes the
height of the photo and the thickness of the photo frame as optional arguments.
`\photo[<photo height >][< frame thickness >]{<photo file name >}`

Keep in mind that the root directory for the .tex file will be `rendercv_output` folder. So for example, if you want to keep your photo in the folder with `Your_Name_CV.yaml` then you should add something like this:
`\photo[3cm][1pt]{../YourPhoto.jpg}`

See more about moderncv documentation [here](https://ctan.math.washington.edu/tex-archive/macros/latex/contrib/moderncv/manual/moderncv_userguide.pdf).

_Note: As this solution only caters `moderncv` theme and I have no idea about other themes then I don't think it's worthy of pull request. Maybe it is, let me know. I just thought it might be good info for some._

sinaatalay:
@kito323 Thank you! All the files inside the theme folder (`moderncv`) will be automatically copied to `rendercv_output` folder. So, putting the photo inside the `moderncv` folder is an option too.

If you would like to contribute the documentation, you can send a pull request to update [the FAQ page](https://docs.rendercv.com/user_guide/faq/). If you're interested, see the [developer guide](https://docs.rendercv.com/developer_guide/). We can update the answer for the other themes later.

sinaatalay:
Profile pictures will be supported in the next release, thanks to #193.

Issue #28, opened 2024-03-04
Ability to add a profile photo


yellowhat:
Thanks for this amazing tool.

It would be great if it would be possible to add a profile picture to the header.

Thanks

sinaatalay:
If someone would like to create a pull request for this feature, it would be greatly appreciated. Thank you for submitting this issue!

PR #174, opened 2024-09-17
Profile picture feture with `moderncv` theme

kito323:
Only for this theme and only optional "add" possibility. Should also include design parameters (height and frame thickness) but didn't know how to add those so that they would have default values.
Also updated user guide FAQ about it.


kito323:
Hope this is how it was intended to be updated. Maybe add `design` parameters for the image too. Currently "hard coded" with height of 3.5cm and no frame. Let me know if it works as intended.

kito323:
Okay I can already see that it fails on ubuntu and macos tests and I know why. Because the path for the image uses different dir delimiters and there are libraries that take care of that but how would that be done in the `Preamble.j2.tex` file or are there any good examples in the codebase that I could follow?

sinaatalay:
Thank you! Sorry for the late response.

I think we shouldn't add this feature to RenderCV itself (built-in `Preamble.j2.tex`), as it won't be available for all the themes. Instead, we should write a step-by-step guide in the FAQ that explains how to add profile pictures to any theme. It should basically say: include a path to a photo in the YAML file and update your `Preamble.j2.tex` accordingly (but more detailed, ofc).

kito323:
Reverted the previous commit and only updated FAQ with the instructions of the simplest implementation. If there are too many details, feel free to adjust it.

kito323:
Maybe also linking the two issues to this PR as partial solution or something ([Ability to add a profile photo #28](https://github.com/sinaatalay/rendercv/issues/28) and [Answer an FAQ: Can I add a profile picture? #74](https://github.com/sinaatalay/rendercv/issues/74)). Not sure if I have permissions to do that...

sinaatalay:
Thank you.

PR #180, opened 2024-10-03
Improve timespan calculation to prevent rounding to 1 year 12 months etc

CaptureTheFrog:
Previous method for calculating timespan allows for rounding up to 12 months even when there are multiple years. "1 year 12 months" isn't correct English, so this fix deals with that overflow as expected.


sinaatalay:
Thank you. I will review it as soon as I can.

sinaatalay:
Thanks again.

PR #193, opened 2024-11-01
Adding support for profile photo

### LINKED ISSUE #28: Ability to add a profile photo

kael-k:
## This PR
Aim to fix #28 by adding the support for the profile picture in a systematic way.

The pull request proposes the following changes:
* [x] add "photo" field in CV model
  * [x] Absolute paths
  * [x] Relative paths
* [x] adding photo in `moderncv` template as suggested in another thread https://github.com/sinaatalay/rendercv/issues/74#issuecomment-2353429721
* [x] adding photo in `sb2nov` template
  * [x] check that the latex code I copypasted and adapted 2 month ago from SO doesn't go bananas with margin changes
* [x] adding photo in the `markdown` template
* [x] adding photo in the `classic` template
* [x] adding photo in the `engineeringresumes` template
* [x] fix tests


rendercv/renderer/templater.py
```diff
@@ -119,11 +119,7 @@ def render_templates(self) -> tuple[str, str, list[tuple[str, list[str], str]]]:
         """
         # Template the preamble, header, and sections:
         preamble = self.template("Preamble")
-        header = (
-            self.template("HeaderWithPhoto")
-            if self.cv.photo
-            else self.template("Header")
-        )
+        header = self.template("Header")
```
kael-k: At the begginning I tought to split header in 2 different templates (header with photo and header without photo), but after seeing that in moderncv it should be implemented in Preamble i reverted it. Also it wasn't a flexible solution, because I was putting as a constraint that the photo had to go in Header

rendercv/data/models/curriculum_vitae.py
```diff
@@ -414,6 +418,13 @@ class CurriculumVitae(RenderCVBaseModelWithExtraKeys):
         alias="sections",
     )
 
+    @pydantic.field_validator("photo")
+    @classmethod
+    def photo_path(cls, value: str | Path, info: pydantic.ValidationInfo) -> Path:
+        """Cast `photo` to Path and make the path absolute"""
+        path = Path(value)
+        return path.absolute()
+
```
kael-k: With this snippet I compute the absolute path of the image. This resolve the problem that relatives path references should change in the j2 template compilation context.

However there is still a problem: relative path is always relative to the cwd, this is the expected behaviour when the parameter is passed by args (ex: --cv.photo=propic.jpg) but I don't think is the correct one if the path is specified in the yaml, because the path should be relative to the yaml and not to the cwd. I think this problem cannot be solved in this PR, as the code merge together YAML and args data before passing it to the pydantic model.

sinaatalay: The line below
https://github.com/sinaatalay/rendercv/blob/9cbff56057ac552dc15ec0e1160c2a740921096a/rendercv/cli/commands.py#L212

changes the current working directory to the parent of the input file. If we move that line above the line below

https://github.com/sinaatalay/rendercv/blob/9cbff56057ac552dc15ec0e1160c2a740921096a/rendercv/cli/commands.py#L196

then, the problem will be resolved and the path will always be relative to the input file path's directory.

rendercv/renderer/templater.py
sinaatalay: Yes, one `Header.j2.tex` file is the way to go; otherwise, there will be code duplication, and we will need to maintain them to be the same.

sinaatalay:
Thank you very much! I will review it when it's done.

rendercv/data/models/curriculum_vitae.py
kael-k: Using os.chdir to change the relative point solve the problem of the relative reference from the YAML but then you have that from the CLI --cv.photo would be relative to the YAML file and not to the cwd (which I think it might be misleading).

kael-k: The only quick solution I see is to explicit the arg "--cv.photo" and, if passed, compute the absolute path from the cwd (idk how typer works, I generally use click, tomorrow I'll take a look to the docs).

sinaatalay: I see your point, but I think it's still fine because we clearly define what that option does, which is overriding the YAML file. So users should expect that `--cv.photo PATH` is equivalent to writing
```yaml
cv:
  photo: PATH
```
which will be equivalent.

kael-k: Makes sense, during next days I'll proceed as you suggested. Thank you for the feedback.

kael-k:
tl;DR: `\usepackage{graphicx}` and `design.font: "Source Sans 3"` in the resume's yaml file do not work together with tinytex, I don't know why.

I implemented the propic for all templates, however, I am facing a strange error with the `classic` template:
the changes I've made are the same for `classic`, `engineeringresume` and `sb2nov`:
1. added `\usepackage{graphicx} % for includegraphics to include profile picture` line in `Preamble.j2.tex` to be able to use `\includegraphics` command
2. changed `Header.j2.tex` to add the profile picture if available

While `engineeringresume` and `sb2nov` work correctly, `classic` does not render correctly and gives me this error in the log file:
```
[...]
LaTeX Font Info:    ... okay on input line 172.

*geometry* driver: auto-detecting
*geometry* detected driver: pdftex
*geometry* verbose mode - [ preamble ] result:
* driver: pdftex
* paper: letterpaper
* layout: <same size as paper>
* layoutoffset:(h,v)=(0.0pt,0.0pt)
* modes: 
* h-part:(L,W,R)=(56.9055pt, 500.484pt, 56.9055pt)
* v-part:(T,H,B)=(56.9055pt, 681.15898pt, 56.9055pt)
* \paperwidth=614.295pt
* \paperheight=794.96999pt
* \textwidth=500.484pt
* \textheight=681.15898pt
* \oddsidemargin=-15.36449pt
* \evensidemargin=-15.36449pt
* \topmargin=-52.36449pt
* \headheight=12.0pt
* \headsep=25.0pt
* \topskip=0.0pt
* \footskip=28.45274pt
* \marginparwidth=65.0pt
* \marginparsep=11.0pt
* \columnsep=4.26773pt
* \skip\footins=9.0pt plus 4.0pt minus 2.0pt
* \hoffset=0.0pt
* \voffset=0.0pt
* \mag=1000
* \@twocolumnfalse
* \@twosidefalse
* \@mparswitchfalse
* \@reversemarginfalse
* (1in=72.27pt=25.4mm, 1cm=28.453pt)


(/Users/kael-k/Documents/rendercv/rendercv/renderer/tinytex-release/TinyTeX/tex
mf-dist/tex/latex/epstopdf-pkg/epstopdf-base.sty
Package: epstopdf-base 2020-01-24 v2.11 Base part for package epstopdf


! LaTeX Error: File `grfext.sty' not found.
```

I tested it both on a macbook air m3 and on a x64 arch linux machine with the same result.
It seems that the error is not because of the changes in `Header.j2.tex`, as if I revert the file and try to render the template with only `\usepackage{graphicx}` I get the same error, which is kinda strange since it works on the other 2 templates. Furthermore, I ran the following tests:
* I tried to render the `classic` template with the `--use-local-latex-command` option and it rendered the resume with the propic successfully
* I tried to take my CV repo, write the template override for the `classic` template with hardcoded path for the image and it rendered the resume with the propic successfully
  * I also tried to rebase my PR on v1.14 (the one I was using in my repo) and then tried to render the resume with the `\usepackage{graphicx}` only, it didn't work


After smashing my head against the PC for 2 hours this is what I got:
* the problem is that the condition in `rendercv/rendercv/renderer/tinytex-release/TinyTeX/texmf-dist/tex/latex/epstopdf-pkg/epstopdf-base.sty:151` seems to be false for `classic` and true for `sb2nov` and `engineeringresumes`, which means that in classic template, tinytex tries to execute L179 `\RequirePackage{grfext}\relax`, which fails.
* The test on my resume with the template override seems to work because the condition on epstopdf-base.sty:151 is false. With that information and some random tests I got that the problem/conflict is caused by the "font part" in Preamble.j2.tex: with `design.font: Latin Modern Sans Serif` in the yaml, the template renders correctly. The problem seems to be with `Source Sans 3`: by setting it manualy in the YAML with `sb2nov` and `engineeringresumes`, you'll get the same error.

@sinaatalay do you have any guesses as to why this happens?

kael-k:
I've found a workaround that I implemented in [a5a43bd](https://github.com/sinaatalay/rendercv/pull/193/commits/a5a43bdde7a208a05cd70be95c9a300c918a6335), I have no idea why it works, but it works

sinaatalay:
Hello @kael-k, can I push some commits before merging this?

kael-k:
> Hello @kael-k, can I push some commits before merging this?

Sure

mekantor:
> I've found a workaround that I implemented in [a5a43bd](https://github.com/sinaatalay/rendercv/pull/193/commits/a5a43bdde7a208a05cd70be95c9a300c918a6335), I have no idea why it works, but it works

I was trying to add graphics (company logo/favicon) and ran into the same problems. What change was this? I can't find it, but I'm glad I wasn't the only one with the issue.

kael-k:
> > I've found a workaround that I implemented in [a5a43bd](https://github.com/sinaatalay/rendercv/pull/193/commits/a5a43bdde7a208a05cd70be95c9a300c918a6335), I have no idea why it works, but it works
> 
> I was trying to add graphics (company logo/favicon) and ran into the same problems. What change was this? I can't find it, but I'm glad I wasn't the only one with the issue.

@mekantor iirc the problem was that the import `graphicx` before  `\usepackage[default, type1]{sourcesanspro}` triggers the import of that `grfext.sty` (because, always iirc, graphicx set some flags that triggers that import in sourcesanspro). The workaround is basically to import graphicx AFTER setting the font.

EDIT: idk why but the commit hash is not available after the rebase I made some days ago, however you can find it by searching the commit with message `fix,template: found a workaround for sourcesanspro and graphicx` (which after the last rebase is https://github.com/rendercv/rendercv/pull/193/commits/960c9d90ceea7d748919d147387925750f22a111)

sinaatalay:
Thank you for your contribution.

PR #198, opened 2024-11-04
Add Markdown Extra extension

anschmieg:
**Feature Description:**
Add Markdown Extra support to the RenderCV renderer. This enhancement would significantly improve the flexibility and styling options for the generated HTML output from CV templates, enabling rendercv to build not just a PDF resumé, but also a personal profile webpage in the same pipeline. 

**Use Cases:**

1. Enhanced Styling: By allowing the use of classes and IDs in Markdown, users can apply custom CSS styles to specific sections of their CVs, resulting in a more personalized and visually appealing presentation.
2. Improved Section Identification: The ability to assign IDs to sections would facilitate easier navigation within the document, especially for longer CVs, improving the overall user experience. This also enables the use of a nav-menu on the page.
3. Better Integration with CSS Frameworks: Many users leverage CSS frameworks that rely on classes for styling. Enabling this feature would allow users to seamlessly integrate their CVs with frameworks like Bootstrap or Tailwind CSS.

**Proposed Solution:**
To implement this feature, I suggest integrating the Markdown Extra extension into the existing rendering pipeline.

### DIFF
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -97,6 +97,7 @@ nav:
 
 markdown_extensions:
   # see https://facelessuser.github.io/pymdown-extensions/extensions/inlinehilite/ for more pymdownx info
+  - extra # Enable the Markdown Extra extensions
   - pymdownx.highlight:
       anchor_linenums: true
       line_spans: __span
--- a/rendercv/renderer/renderer.py
+++ b/rendercv/renderer/renderer.py
@@ -320,9 +320,11 @@ def render_an_html_from_markdown(markdown_file_path: pathlib.Path) -> pathlib.Pa
     if not markdown_file_path.is_file():
         raise FileNotFoundError(f"The file {markdown_file_path} doesn't exist!")
 
-    # Convert the markdown file to HTML:
+    # Convert the markdown file to HTML, supporting Markdown Extra:
     markdown_text = markdown_file_path.read_text(encoding="utf-8")
-    html_body = markdown.markdown(markdown_text, extensions=["toc", "extra"]) # uses the TOC and Markdown Extra extensions
+    html_body = markdown.markdown(
+        markdown_text, extensions=["extra"]
+    )
 
     # Get the title of the markdown content:
     title = re.search(r"# (.*)\n", markdown_text)
--- a/rendercv/renderer/renderer.py
+++ b/rendercv/renderer/renderer.py
@@ -322,7 +322,7 @@ def render_an_html_from_markdown(markdown_file_path: pathlib.Path) -> pathlib.Pa
 
     # Convert the markdown file to HTML:
     markdown_text = markdown_file_path.read_text(encoding="utf-8")
-    html_body = markdown.markdown(markdown_text)
+    html_body = markdown.markdown(markdown_text, extensions=["toc", "extra"]) # uses the TOC and Markdown Extra extensions
 
     # Get the title of the markdown content:
     title = re.search(r"# (.*)\n", markdown_text)
--- a/rendercv/themes/main.j2.html
+++ b/rendercv/themes/main.j2.html
@@ -12,6 +12,9 @@
         href="https://cdnjs.cloudflare.com/ajax/libs/github-markdown-css/5.5.1/github-markdown-light.min.css"
         integrity="sha512-Pmhg2i/F7+5+7SsdoUqKeH7UAZoVMYb1sxGOoJ0jWXAEHP0XV2H4CITyK267eHWp2jpj7rtqWNkmEOw1tNyYpg=="
         crossorigin="anonymous" referrerpolicy="no-referrer" />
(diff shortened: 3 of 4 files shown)


anschmieg:
okay so even though I basically changed one line I messed something up. Maybe someone with more experience knows what to do...?

anschmieg:
Previous commit was faulty (because I added a faulty call to the TOC extension), this one works on my end and passes the tests. One issue that remains - at least for my desired use case - is that ids and classes can only be added to certain elements according to the Markdown Extra spec. It is what it is. 

BTW this is my first contribution to someone else's project. I hope that everything is correct and this creates value added.

sinaatalay:
Thank you very much. I will have a look at my earliest opportunity.

anschmieg:
Great, no worries if it takes you a while. I am super glad I found this project, since I was working on something similar on a much simpler level before. So thank you for all your amazing work! 

I realized another thing about the behavior of markdown with HTML that I wanted to document for reference, just in case.

**Note for usage**
When a markdown source is wrapped in an HTML element, like 
`<div> I _really_ like the **syntax** of Markdown</div>`
it gets rendered as the literal
> I \_really\_ like the \*\*syntax\*\* of Markdown

To render the markdown within the block, add the `markdown='1'` attribute as follows:
`<div markdown='1'> I _really_ like the **syntax** of Markdown</div>`
and it will be rendered as 
> I _really_ like the **syntax** of Markdown

If you are using the markdown templates to structure the HTML output, you may want to use wrappers and will hence face that issue. With Markdown Extra enabled, you can circumvent this by adding the `markdown' attribute to the immediate parent HTML block's tag (of the markdown element).

sinaatalay:
Could you please give an example YAML file that showcases the capability that isn't possible without this extension?

anschmieg:
I'm not quite sure I understand what you mean. Let's say e.g. my YAML file contains

```yaml
cv:
[...]
  email: me@example.com
  phone: tel:+441234567890
  website: https://example.com/
  social_networks:
    - network: LinkedIn
      username: me
    - network: Github
      username: notme
```

**For use case 1 and 3 mentioned in the first comment:**
now on the HTML page, I may want to, say, move the social networks to a sticky block on the right side. Without using classes, I either have to modify the template and figure out how to separate everything i need to, or use weird css selectors like `ul li a[href*="http"]`, which would make it somewhat unstable if anything on the CV changes. Having predefined classes would all make all of that redundant without adding any complexity to the structure of the document itself, i.e. without (notably?) affecting load speeds etc.

**For use case 2:**
Currently, e.g. to create a navigation menu or to directly link to a section, the best solution I could come up with is probably adding something like this to the template:

```JavaScript
document.querySelectorAll('h1').forEach((heading) => {
    // Create a unique ID from the heading text and a random number
    const id = heading.textContent.someFormatting() + Math.random().someFormatting()
    heading.id = id;
});
```

If, instead, my YAML could contain (I didn't actually check this with rendercv because I'm on another device, but I think the syntax should be fine)
```yaml
experience {#section-experience}:
  - [...]
```

I could have my menu and/or direct links link to `https://personal.rendercv.app/en.html#section-experience` or whatever.

---

Does this clarify it? As previously mentioned, I am talking about using the HTML for an actual webpage, not for Grammarly. 

If this is not what you asked for, just let me know please.

sinaatalay:
I don't know much about markdown extensions. So, I need an example to understand this PR. 

Could you provide a YAML file that uses the new features this PR implements so that we can checkout to this pull request, try the YAML file, and see how it works?

anschmieg:
Now I understand. What I meant was a YAML input like

```yaml
sections:
  education {#education}:
    - institution: University of Pennsylvania
      area: Computer Science {#bachelors}
      degree: BS
      start_date: 2000-09
      end_date: 2005-05
      highlights:
        - 'GPA: 3.9/4.0 ([a link to somewhere](https://example.com))'
        - '**Coursework:** Computer Architecture, Comparison
          of Learning Algorithms, Computational Theory'
```

Upon preparing the example, I noticed that this input results in a perfectly fine HTML output, which was the goal: 

```html
<h1 id="education">Education</h1>
<h2 id="bachelors">University of Pennsylvania, BS in Computer Science</h2>
<ul>
  <li>Sept 2000 – May 2005</li>
  <li>GPA: 3.9/4.0 (<a href="https://example.com">a link to somewhere</a>)</li>
  <li><strong>Coursework:</strong> Computer Architecture, Comparison of Learning Algorithms, Computational Theory</li>
</ul>
```

However, stupidly, I never considered that it would also affect the LaTeX output and therefore the PDF. Since that doesn't use Markdown (and hence Markdown Extra), it prints the full expressions as `Education {#education}` etc. 

In order to implement this properly, I will have to find a way to exclude these Markdown instructions from the conversion to LaTeX. I will have a look at that at another time. It is probably best to reject this PR and I will open a new one when I have it ready.

Sorry for the turbulence, looks like I'm having some newbie trouble.

sinaatalay:
No need to be sorry. Thank you very much for the ideas! Feel free to comment under this PR if anything else comes to your mind.

PR #199, opened 2024-11-06
Add support for LaTeX math expressions in HTML output

anschmieg:
**Issue Description**
Currently, the HTML output displays e.g. the example entry
```
Built an app to compute the similarity of all methods
in a codebase, reducing the time from $$\mathcal{O}(n^2)$$
to $$\mathcal{O}(n \log n)$$
```
as a literal, without rendering the math expression. 

**Suggested change**
By adding Mathjax to the document via the Cloudflare CDN, LaTeX and LaTeX math expressions get parsed correctly if the CDN is reachable.


anschmieg:
I messed up the commit history, closing this PR and creating a new one.

PR #200, opened 2024-11-06
Add support for LaTeX math expressions

anschmieg:
**Issue Description**
Currently, the HTML output displays e.g. the example entry
```
Built an app to compute the similarity of all methods
in a codebase, reducing the time from $$\mathcal{O}(n^2)$$
to $$\mathcal{O}(n \log n)$$
```
as a literal, without rendering the math expression. 

**Suggested change**
By adding Mathjax to the document via the Cloudflare CDN, LaTeX and LaTeX math expressions get parsed correctly if the CDN is reachable. 

--- 

I hope I didn't screw up the branches or the commit history again... still working those out 😄

### DIFF
--- a/rendercv/themes/main.j2.html
+++ b/rendercv/themes/main.j2.html
@@ -12,6 +12,9 @@
         href="https://cdnjs.cloudflare.com/ajax/libs/github-markdown-css/5.5.1/github-markdown-light.min.css"
         integrity="sha512-Pmhg2i/F7+5+7SsdoUqKeH7UAZoVMYb1sxGOoJ0jWXAEHP0XV2H4CITyK267eHWp2jpj7rtqWNkmEOw1tNyYpg=="
         crossorigin="anonymous" referrerpolicy="no-referrer" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/3.2.2/es5/tex-mml-chtml.min.js"
+        integrity="sha512-6FaAxxHuKuzaGHWnV00ftWqP3luSBRSopnNAA2RvQH1fOfnF/A1wOfiUWF7cLIOFcfb1dEhXwo5VG3DAisocRw=="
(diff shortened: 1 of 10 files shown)


anschmieg:
_Original comment:_ Since there is still an AssertionError, I assume the HTML output doesn't match the expected result because of the added tag?

**Edit: committed modification of the relevant tests**

sinaatalay:
Thank you, I haven't thought about this before! I will merge this. Could you use [KaTeX](https://katex.org) instead? It's much faster than MathJax and it's what we use on [docs.rendercv.com](https://docs.rendercv.com).

anschmieg:
> I haven't thought about this before

I guess it is a pretty niche use case to need math expressions in your CV. Considering that, it might make sense to lazy-load the lib to avoid unnecessary bloat for everyone else. However, I am not aware of that being possible without using another library, which would kinda defeat the purpose. Thoughts?

> Could you use [KaTeX](https://katex.org) instead? 

Cool, I didn't know that one! I suppose it should be a drop-in replacement... I would still embed it as a JsDeliver reference though, for simplicity's sake. Unless you have a better idea. Shall I update the PR?

sinaatalay:
KaTeX is very lightweight, so there is no problem at all with including it in the HTML. If you could update this PR for KaTeX, it will be ready to be merged.

anschmieg:
_"I am once again asking for your technical support"_

The test I added in [5d7e98f](https://github.com/sinaatalay/rendercv/pull/200/commits/5d7e98f621cfb184dcc316763cb342d34f630ca8) appears to have been a legacy test case. I reverted the change, both the feature and the tests should be fine now. I wouldn't mind a review though.

sinaatalay:
Thank you! It was not working but the fix was quick ([details](https://github.com/rendercv/rendercv/pull/200/commits/3142fb6b50362c8890bdc864ada3147d8940918f)).

PR #209, opened 2024-11-16
Solution to Issue# 148

MTaimoor99:
I have moved `last_update_style` and `page_numbering_style` fields to `locale_catalog` field in `schema.json`


sinaatalay:
Thank you! This is not something related to `schema.json` file but the source code of the RenderCV. `schema.json` file is created automatically from the source code as described [here](https://docs.rendercv.com/developer_guide/writing_documentation/#updating-the-json-schema-schemajson).

Issue #210, opened 2024-11-16
Add summary field to entries


kiyose:
Many times entries should have a summary field in addition to highlights.  

For example:
```
   projects:
      - name: Multi-User Drawing Tool
        summary:
          - The Multi-User Drawing Tool has been released to tens of users saving thousands of microseconds worth of effort.
        highlights:
          - Developed an electronic classroom where multiple users
            can simultaneously view and draw on a "chalkboard"
```

This is a field that is defined in JSONResume schema as well and seems to be the one area of mismatch.  

updated: PR #211 adds .summary to the entities and templates

tobybaratta:
I was wondering if it was possible for this to get merged in and released? This would be awesome to have as a feature (and super wanted!).

PR #211, opened 2024-11-16
Adds summary field to data model and templates

kiyose:
This fixes issue #210 by adding a summary field to entry types.

For example:
```  
projects:
      - name: Multi-User Drawing Tool
        summary:
          - The Multi-User Drawing Tool has been released to tens of users saving thousands of microseconds worth of effort.
        highlights:
          - Developed an electronic classroom where multiple users
            can simultaneously view and draw on a "chalkboard"
```

The only change in functionality is in the data model here `rendercv/data/models/entry_types.py`.  The rest of the files are updating the templates and tests.


sinaatalay:
Could you please provide the screenshots of what `summary` key creates exactly?

kiyose:
Here are two entries before from before this PR.
without_summary


and here are the same entries after adding the summary.
with_summary

dyliu0312:
Thank you, guys.
This is exactly the kind of functionality I was hoping for, as long as it also works in the _Publication_ section

sinaatalay:
Thank you for the pull request!

I think it would be the best to not merge this.

We are planning to transition to Typst soon, and each design update like this increases the complexity of maintenance. We need to ensure that design (LaTeX templates) works as expected with various combinations of `design` settings, which requires extensive manual testing. Currently, RenderCV runs robustly with its LaTeX templates, and we wouldn't want to break it right before we switch to Typst. Typst will require a lot of work. We can implement these in Typst.

For those interested in using these keys in RenderCV:

It is already quite simple to do so. That's why we allow users to [use arbitrary keys in their YAML files](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#using-arbitrary-keys) and [override the default LaTeX templates](https://docs.rendercv.com/user_guide/#overriding-built-in-themes). Check out the links for details.

Issue #109, opened 2024-06-11
Remove Twitter social network since it has been deprecated and renamed to X


TheColorDarkBlue:
This rebrand is almost a year old. Twitter profiles automatically redirect from **twitter.com** to **x.com**. 

https://www.theverge.com/2024/5/17/23829098/twitter-x-com-url-links-switch

#101 #108 #107

sinaatalay:
Thank you. Twitter has been removed from RenderCV in v1.11.

PR #212, opened 2024-11-17
Twitter social

mrexodia:
I see it was removed after #109, but there is currently no replacement. I'd rather see an 'outdated' Twitter icon than not being able to link my X account at all.


sinaatalay:
Thank you. Twitter and its icon don't exist anymore. Therefore, I think including it would be wrong. I think we need to go for the X icon somehow.

docs/developer_guide/faq.md
```diff
@@ -14,5 +14,7 @@ Then, the tests should be implemented for the new social network with the follow
 
 1. Go to `tests/test_data.py` and update `test_social_network_url` accordingly, i.e., add a new `(network, username, expected_url)` tuple to the `pytest.mark.parametrize` decorator.
 2. Go to `tests/conftest.py` and add the new social network to `rendercv_filled_curriculum_vitae_data_model`.
-3. Set `update_testdata` to `True` in `conftest.py` and run the tests to update the `testdata` folder.
+3. Set `update_testdata` to `True` in `conftest.py` and [run the tests](./testing.md) to update the `testdata` folder.
 4. Review the updated `testdata` folder manually to ensure everything works as expected. Then, set `update_testdata` to `False` and push the changes.
+
+Finally, add the new social network to `schema.json` to finish the integration of the new social network.
```
sinaatalay, line 20: `schema.json` is generated automatically. See [here](https://docs.rendercv.com/developer_guide/writing_documentation/#updating-the-json-schema-schemajson).

docs/developer_guide/faq.md
```diff
@@ -14,5 +14,7 @@ Then, the tests should be implemented for the new social network with the follow
 
 1. Go to `tests/test_data.py` and update `test_social_network_url` accordingly, i.e., add a new `(network, username, expected_url)` tuple to the `pytest.mark.parametrize` decorator.
 2. Go to `tests/conftest.py` and add the new social network to `rendercv_filled_curriculum_vitae_data_model`.
-3. Set `update_testdata` to `True` in `conftest.py` and run the tests to update the `testdata` folder.
+3. Set `update_testdata` to `True` in `conftest.py` and [run the tests](./testing.md) to update the `testdata` folder.
```
sinaatalay, line 17: If you could send this line as a different PR, I would like to merge it.

mrexodia:
> Thank you. Twitter and its icon don't exist anymore. Therefore, I think including it would be wrong. I think we need to go for the X icon somehow.

I do not understand this response. The Twitter icon still exists in fontawesome5 and will exist forever. As it stands there is no way to link an X account, so in my view it is better to use the old icon?

sinaatalay:
Hi, I'm sorry, I didn't see this response. What I meant was: Twitter as a platform doesn't exist, right? I don't think supporting something that doesn't exist is good.

We plan to switch to [Typst](https://typst.app) soon. We can add X support after we switch to Typst.

mrexodia:
Yeah I also agree that Twitter no longer exists, but now you are essentially removing a feature (displaying your Twitter/X on your CV) until switching to a new backend. Isn't is better to show the Twitter icon and link people to `x.com` than to not be able to display your X profile _at all_?

sinaatalay:
I honestly believe it's better to not have X social network support than to mislead users to use the wrong icon in their CVs. Using the wrong icon is just as problematic as having a spelling mistake in a CV, in my opinion.  One of the goals of RenderCV is to help people not make mistakes with their CVs and content.

Also, the `Header.j2.tex` file [can be overridden](https://docs.rendercv.com/user_guide/#overriding-built-in-themes) by users to use any icon and social network in the end.

PR #220, opened 2024-11-18
Allow for seperate design and rendercv settings from the CV data

kiyose:
This PR allows users to separate their rendering settings from the theme design settings from their CV data while not requiring them to do so.

It adds two flags to the cli command
```
│ --design-path                     -d           TEXT  Use the design settings located at this path. [default: None]                                                                                                                                                          
│ --rendercv-settings-path   -s           TEXT  Use the rendercv settings located at this path. [default: None]
```

This allows for the same settings to be applied to different CV data without having to copy and paste as well as quickly changing the theme used without having to update the CV data file.


sinaatalay:
Thank you! This was a very good idea. I've made some changes in here: https://github.com/rendercv/rendercv/pull/220/commits/af9990fde5db075fa33265e2d6a444202afaf58f

PR #235, opened 2024-11-30
Update faq.md

guruor:
As suggested in #150, Mentioned jsonresume-to-rendercv in FAQ for JSON Resume schema users.


sinaatalay:
Thank you!

PR #239, opened 2024-12-03
Add support for multiple positions at the same employer

mekantor:
When being with a single employer for a long time and progressing through titles, it may be desirable to show this.
Use the `positions` list to implement this, with the properties `name`, `start_date`, and `end_date`.
When templates detect `positions`, they will ignore `position` and prefer to iterate over this list.


sinaatalay:
Thank you! We recently decided to switch to Typst, therefore I think we shouldn't merge this. I am copying and pasting my response to another PR (#211), as it's for the same reason:

_I think it would be the best to not merge this._

_We are planning to transition to Typst soon, and each design update like this increases the complexity of maintenance. We need to ensure that design (LaTeX templates) works as expected with various combinations of design settings, which requires extensive manual testing. Currently, RenderCV runs robustly with its LaTeX templates, and we wouldn't want to break it right before we switch to Typst. Typst will require a lot of work. We can implement these in Typst._

_For those interested in using these keys in RenderCV:_

_It is already quite simple to do so. That's why we allow users to [use arbitrary keys in their YAML files](https://docs.rendercv.com/user_guide/structure_of_the_yaml_input_file/#using-arbitrary-keys) and [override the default LaTeX templates](https://docs.rendercv.com/user_guide/#overriding-built-in-themes). Check out the links for details._

PR #249, opened 2024-12-10
Update commands.py

S-Avalloni:
Remove bug where 'design' was not recognized as a valid key in locals() by declaring cli_render_arguments explicitly.
Cli was installed by  `pip install rendercv` on python 3.11.2 with 24.0.
Full error on image if fix doesn't fits intended purpose

imagen

### DIFF
--- a/rendercv/cli/commands.py
+++ b/rendercv/cli/commands.py
@@ -172,21 +172,21 @@ def cli_command_render(
     from . import utilities as u
 
     cli_render_arguments = {
-        'design': design,
-        'locale_catalog': locale_catalog,
-        'rendercv_settings': rendercv_settings,
-        'use_local_latex_command': use_local_latex_command,
-        'output_folder_name': output_folder_name,
-        'latex_path': latex_path,
-        'pdf_path': pdf_path,
-        'markdown_path': markdown_path,
-        'html_path': html_path,
-        'png_path': png_path,
-        'dont_generate_markdown': dont_generate_markdown,
-        'dont_generate_html': dont_generate_html,
-        'dont_generate_png': dont_generate_png,
-        'watch': watch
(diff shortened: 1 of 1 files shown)


sinaatalay:
Thank you! I think we should find a solution and stick to the old code. I am surprised about the issue, it was being tested. Could you please share the CLI command you used?

S-Avalloni:
The issue also surprised me, I tried to find a solution that was more closely related to your code but was unable to find one, locals() did have 'design' as a valid key so maybe it had something to do with the interaction between locals() and the list comprehension.

I was using `rendercv render Santiago_Avalloni_CV.yaml` without any additional arguments and the change was made based on the expected output of the `argument_names` variable because default cli arguments are the same always (also because of this same reason it's not necessary to compute it every time)

S-Avalloni:
I figured out the problem, the thing is that locals() is a dict containing all variables **of the current scope** so when you are doing a dict comprehension locals() is not the full dict, but a dict containing only the keys 'name' and 'argument_names'. It's difficult to see explicitly but if you do

``` py
var = 1
a = [locals() for i in range(1)]
```

`a` is a list containing only `{'.0': <range_iterator object at 0x000001B5D824BD30>, 'i': 1}` and not `var`

sinaatalay:
The original code works for me. Could you please share the CLI command that caused this error? I cannot reproduce it.

EDIT: I reproduced it. It's happening with Python 3.10 and 3.11. Thank you, this will be fixed.

sinaatalay:
I pushed the fix to this PR (https://github.com/rendercv/rendercv/pull/249/commits/4fd6dc33f2c0881887e8be90876e1d2b381f286b).

Thank you for your contribution.

PR #260, opened 2024-12-16
WIP: Company logo

mekantor:
Purpose: Show a company logo next to the name.

Method: Include a file in the same path as the YML content. Specify as `logo: basename` and it will select a file with that name and a supported extension. Supported formats for latex: `jpg`, `png`, `pdf`, `eps`.

TBD: need to add markdown and HTML output

Blocked by https://github.com/rendercv/rendercv/pull/193 and https://github.com/rendercv/rendercv/pull/239


sinaatalay:
Thank you for the PR.

It's still draft but I have two concerns about this:

- Is it okay to use companies logos in our CVs? Copyright issues, etc. I see a lot of discussion on the internet that says it's not recommended to use company logos. If it's not, we shouldnt have this feature in RenderCV. RenderCV shouldn't mislead people to do mistakes in their CVs.
- We are switcihing to Typst soon, so it would be the best to wait for that first.

mekantor:
> Thank you for the PR.
> 
> It's still draft but I have two concerns about this:
> 
> * Is it okay to use companies logos in our CVs? Copyright issues, etc. I see a lot of discussion on the internet that says it's not recommended to use company logos. If it's not, we shouldnt have this feature in RenderCV. RenderCV shouldn't mislead people to do mistakes in their CVs.
> * We are switcihing to Typst soon, so it would be the best to wait for that first.

This is a good point, I suppose the only counterpoint I can offer os that company logos are routinely on hiring sites, and LinkedIn shows logos on work history.

PR #269, opened 2024-12-25
Added markdown_to_latex as a Jinja filter

maharjun:
As described, it makes sense to have this as a filter because nested elements are not connected from markdown to latex when using custom themes with different keys.


sinaatalay:
Thank you.

PR #279, opened 2024-12-30
References

artnim:
Add an entry type for references.


sinaatalay:
Thanks. Could you share a screenshot of the new entry type? Does it have a look that cannot be achieved with one of the seven entry types?

artnim:
Here are some screenshots from my local examples folder.
Screenshot 2024-12-31 at 18 13 02
Screenshot 2024-12-31 at 18 13 35
Screenshot 2024-12-31 at 18 14 28
Screenshot 2024-12-31 at 18 15 04

sinaatalay:
Thank you. We switched to Typst, so I need to close this PR.

Also, this look can be achieved with NormalEntry, so we don't need a new entry type for it. RenderCV's entry types are not data objects (even though they look like data storage objects). They are "visual objects" with distinct looks in the PDF. That's why I asked for a screenshot before. I was wondering if this entry was adding a new visual look that couldn't be achieved with the other entry types.

PR #287, opened 2025-01-06
fix: build new dict when bolding to avoid shared references

### LINKED ISSUE #286: new bold_keywords setting doesn't work with multiple keywords
Rendering a YAML specifying keywords under `rendercv_settings.bold_keywords` should surround occurrences of those keywords with double asterisks, bolding those keywords in the result PDF. But, using v1.18 of the CLI, specifying two or more keywords to bold will result in unexpected behavior.

## To reproduce
Render the following YAML with `rendercv render debug.yaml -use pdflatex`

```yaml
cv:
  name: Ben Juntilla
  sections:
    experience:
      - company: Circle K
        position: Snack Eater
        highlights:
          - bananas apples pears

design:
  theme: engineeringresumes
  font: Source Sans 3

rendercv_settings:
  bold_keywords:
    - bananas
    - apples
    - pears
```

## Screenshots
**Bolding one keyword:**
```yaml
    - bananas
    # - apples
    # - pears
```
image
(Works!)

**Bolding two keywords:**
```yaml
    - bananas
    - apples
    # - pears
```
image
(Not so much)

**Bolding three keywords:**
```yaml
    - bananas
    - apples
    - pears
```
image
(Not at all)

**Bolding four keywords (also added an additional word in the highlights):**
```yaml
    - bananas
    - apples
    - pears
    - oranges
```
image

## Possible root cause
I think it's because `make_given_keywords_bold_in_a_dictionary` in `utilities.py` creates a shallow copy with `dictionary.copy()`, resulting in shared references between the new and old dictionaries. I already have a fix that works on my end, so I'll open a PR soon.

benjuntilla:
resolves #286 by fixing an issue in `make_given_keywords_bold_in_a_dictionary` where a shallow copy (`dictionary.copy()`) could lead to shared references between old and new dicts, causing inconsistent bolding when more than one keyword is specified. The updated function constructs a new dictionary so that no references to old data remain.

### DIFF
--- a/rendercv/cli/utilities.py
+++ b/rendercv/cli/utilities.py
@@ -319,23 +319,32 @@ def make_given_keywords_bold_in_a_dictionary(
     Returns:
         The dictionary with the given keywords bold.
     """
-    new_dictionary = dictionary.copy()
-    for keyword in keywords:
-        for key, value in dictionary.items():
-            if isinstance(value, str):
-                new_dictionary[key] = value.replace(keyword, f"**{keyword}**")
-            elif isinstance(value, dict):
-                new_dictionary[key] = make_given_keywords_bold_in_a_dictionary(
-                    value, keywords
-                )
-            elif isinstance(value, list):
-                for i, item in enumerate(value):
-                    if isinstance(item, str):
-                        new_dictionary[key][i] = item.replace(keyword, f"**{keyword}**")
-                    elif isinstance(item, dict):
-                        new_dictionary[key][i] = (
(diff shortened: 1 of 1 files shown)


sinaatalay:
Thanks a lot for pointing this out. However, I was working on v2 [here](https://github.com/rendercv/rendercv/pull/271) and was already planning to move this from `cli` to `data`, as it has nothing to do with CLI. Therefore, I won't be able to merge this.

PR #299, opened 2025-01-09
Improve packaging version comparison

### LINKED ISSUE #298: Version comparison warning even if newer version installed
Changing version in `rendeercv::__init__.py` causes a warning message, even though the version is newer than published:

```
$ rendercv --version
A new version of RenderCV is available! You are using v2.1.dev0, and the latest version is v2.0.
```

lczyk:
Fixes #298.

Also uses the official [packaging](https://packaging.pypa.io/en/stable/) package to compare versions form pypi, which allows properly comparing [PEP 440](https://peps.python.org/pep-0440/) versions, like the dev version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ dependencies = [
     'pydantic==2.9.2',              # to validate and parse the input file
     'pycountry==24.6.1',            # for ISO 639-3 validation
     'pydantic-extra-types==2.10.1', # to validate some extra types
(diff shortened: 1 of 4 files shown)


sinaatalay (approved):
Thank you very much. I made small changes [here](https://github.com/rendercv/rendercv/pull/299/commits/3c2ac6f82da5925f11657ca3d91d3c701fa8f759).

PR #308, opened 2025-01-14
Update README.md

jaehho:
**Fix Blog Post hyperlink**

The old hyperlink is sending to a broken link that does not exist anymore.
Replacing with the only Blog Post I see on the website


sinaatalay:
Thank you for pointing this out. I have instead updated the blog post's URL as I had forgotten to do so earlier.

PR #310, opened 2025-01-17
Fix path issues in `rendercv_settings` and CLI

### LINKED ISSUE #272: External YAML breaks with spaces in path

george-gca:
Fixes #272. That code fails even without spaces in the path.

### DIFF
--- a/rendercv/cli/commands.py
+++ b/rendercv/cli/commands.py
@@ -45,7 +45,7 @@ app = typer.Typer(
 def cli_command_render(
     input_file_name: Annotated[str, typer.Argument(help="The YAML input file.")],
     design: Annotated[
-        Optional[str],
+        Optional[pathlib.Path],
         typer.Option(
             "--design",
             "-d",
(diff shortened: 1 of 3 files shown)


sinaatalay:
Thank you for the PR, but the root of the issue was in `data.models.rendercv_settings.py`. The types of `design`, `rendercv_settings`, and `locale` was `str` instead of `Path`. See #312.

PR #321, opened 2025-01-23
Link docker and pypi releases

### LINKED ISSUE #320: Race condition: docker and pypi workflows
The docker and pypi workflows are both triggered at the same time: when a github release is created.

The docker workflow installs the most recent version of the package from pypi. The docker build is faster than the pypi release, resulting in the previous version of the package being installed.

This is seen in both the v2 and v1.18 versions of the docker image (I did not check earlier versions).

mheguy:
Closes #320

This resolves the race condition between the pypi and docker workflows when triggered by a github release by combining the two workflows and having docker depend on pypi.

I added the "release" environment to the docker job in order to match what was in the pypi job, though I don't think you actually use deployments and/or environments.

While the changes are straightforward, I don't know of a way to test for issues until the next release.


sinaatalay:
Thank you!

mheguy:
@sinaatalay the release is failing.
That's on me, I renamed the file without realising it would break things.
On Pypi, you'll need to set the correct workflow name.
https://docs.pypi.org/trusted-publishers/adding-a-publisher/

sinaatalay:
Ahh, right. Thank you, I was trying to figure out why.

PR #365, opened 2025-02-22
Change code creating rendercv project in index.md

jacekkala:
When copying and running:

```
rendercv "Your Full Name"
```

from **User Guide**, the following error is raised:

rendercv : The term 'rendercv' is not recognized as the name of a cmdlet, function, script file, or operable program.

However, when used this code:

```
python -m rendercv "Your Full Name"
```

everything works perfectly. I used PowerShell, of course rendercv was installed. I am wondering if the second version (with python -m) would be transferable to all systems and avoiding "rendercv not recognized" errors.

### DIFF
--- a/docs/user_guide/index.md
+++ b/docs/user_guide/index.md
@@ -17,7 +17,7 @@ pip install "rendercv[full]"
 To get started, navigate to the directory where you want to create your CV and run the command below to create the input files.
 
 ```bash
-rendercv new "Your Full Name"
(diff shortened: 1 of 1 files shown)


marcorichetta:
Most likely you need to [activate your virtual environment](https://docs.python.org/3/library/venv.html#how-venvs-work) (Ex `source .venv/bin/activate`).
IMO it would be better adding the virtualenv activation step before installing `rendercv`

Issue #362, opened 2025-02-17
Spelling mistake "type: with-parial-line"


amaanansari:
**Describe the bug**
The interpreter expects a misspelled option:
```
design:
  section_titles:
    type: with-parial-line # this value is expected, but it should be "with-partial-line"
```

**To Reproduce**
rendercv==2.2

**Screenshots**

![Image](https://github.com/user-attachments/assets/ba6a9dec-b4cc-42fe-9824-955558492447)

NeodarZ:
This issue is not resolved?

Like there is already a closed issue from the day before this one is posted: https://github.com/rendercv/rendercv/issues/349

thomascamminady:
This error still exists for me as well today, using `rendercv==2.2`.

clintval:
I also have the same issue on `2.2` and notice there are still references to the mispelling:

```console
❯ git grep parial
rendercv/__init__.py:_parial_install_error_message = (
rendercv/cli/__init__.py:    from .. import _parial_install_error_message
rendercv/cli/__init__.py:    print(_parial_install_error_message)  # noqa: T201
rendercv/renderer/renderer.py:                from .. import _parial_install_error_message
rendercv/renderer/renderer.py:                raise ImportError(_parial_install_error_message) from e
rendercv/renderer/renderer.py:        from .. import _parial_install_error_message
rendercv/renderer/renderer.py:        raise ImportError(_parial_install_error_message) from e
(END)
```

S-Avalloni:
This is still present, the schema.json doesn't accept `with-parial-line` but as is shown above that's the expected input and the one that works as intended

PowerUser64:
Looks like this was resolved in v2.3

Issue #349, opened 2025-02-09
maybe spelling error: with-parial-line


tsvikas:
there is a class/type named `with-parial-line`. is it a spelling error? 

if so, a quick search and replace can fix it (will be happy to PR if needed)

sinaatalay:
Yes, it's a spelling mistake. Thank you!

sinaatalay:
Hello, sorry, I need to update this myself as we prepare an update for the web app. Thank you for the issue again.

ShoGinn:
I see the issue has been resolved; what is your release timeline to push this to prod?

I currently am using UV to manage this with building main.. thanks in advance @sinaatalay

sinaatalay:
Sorry, I haven't been doing a good job maintaining RenderCV in the past weeks. I will return very soon, merge everything, and release a new version (hopefully in a couple of weeks).

PR #380, opened 2025-03-19
Search and replace 'parial' with 'partial'.

thomascamminady:
Should address https://github.com/rendercv/rendercv/issues/362


ShoGinn:
> Should address #362

I disagree. 

This will not address #362 as this is just to fix the variables for error messages.

thomascamminady:
Thanks for having a closer look and sorry for the false assumption on my end. Shall I rename the PR to something more appropriate and leave it with fixing this typo in this PR?

ShoGinn:
> Thanks for having a closer look and sorry for the false assumption on my end. Shall I rename the PR to something more appropriate and leave it with fixing this typo in this PR?

Certainly; I have discovered that being more direct with language facilitates easier comprehension.

This remains a valid issue, and the misspellings should be corrected!

The absence of a fix in #362 is attributed to the fact that no version has been released since issue #349.

sinaatalay:
Thank you.

PR #395, opened 2025-03-26
fixed wrong left-sum-right-margin in connections

tsvikas:
`left-sum-right-margin` was set to the correct size: left + right, i.e. `page.margin * 2`
the wrong value created premature newlines in the connection box. now it can span the full line, saving on vertical space.

### DIFF
--- a/src/rendercv/themes/classic/Preamble.j2.typ
+++ b/src/rendercv/themes/classic/Preamble.j2.typ
@@ -330,7 +330,7 @@
     if type(page.margin) == "dictionary" {
       left-sum-right-margin = page.margin.left + page.margin.right
     } else {
-      left-sum-right-margin = page.margin * 4
+      left-sum-right-margin = page.margin * 2
     }
 
     let ending-index = starting-index + 1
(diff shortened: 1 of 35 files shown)


amarvin:
Nice find! I think to get the tests passing, you just need to update the expected PDF files with the updated margins.

amarvin:
Nice progress! Looks like only this test is failing now: `tests/test_renderer.py::test_copy_theme_files_to_output_directory_custom_theme`.

tsvikas:
Ready for review now. @sinaatalay would you want to take a look? :)

sinaatalay:
Thanks a lot. I will merge it as soon as possible.

sinaatalay:
I apologize for the very late response. I'm back to maintaining RenderCV.

I was going to merge this, but I realized the change breaks the logic:
image

The right side shows the new test data, and the left side shows the old one. I haven't examined the logic we have there closely yet, but changing the multiplier from *4 to *2 causes incorrect line breaks. We need to understand why this happens before proceeding.

Issue #407, opened 2025-04-09
Allow new fields in cv section


nan0m:
I need more flexibility in what goes in the top section. 
Sometimes a second website would be great. While other times I e.g. wanna mention my birthdate or nationality. 
I think it should be possible to add variable links there.

sinaatalay:
This is now possible with `cv.custom_connections` field (see [docs](https://docs.rendercv.com/user_guide/yaml_input_structure/cv/)) in v2.4.

nan0m:
Thank you so much! @sinaatalay

Issue #383, opened 2025-03-20
Feedback from professionals


niccokunzmann:
I had feedback from professionals: for Austria, you also need:

- date of birth
- nationality

They should probably go just below the name:

```yaml
cv:
  birth: 1990-11-23
  nationality: German
```

One could also solve this with a free text.

Thanks for this software!

grigorishat:
> I had feedback from professionals: for Austria, you also need:
> 
> * date of birth
> * nationality
> 
> They should probably go just below the name:
> 
> cv:
>   birth: 1990-11-23
>   nationality: German
> One could also solve this with a free text.
> 
> Thanks for this software!

To include birth and nationality, I added the following lines

```PYTHON
        if self.birth is not None:
            connections.append(
                {
                    "typst_icon": "cake-candles",
                    "url": None,
                    "clean_url": None,
                    "placeholder": self.birth,
                }
            )

        if self.nationality is not None:
            connections.append(
                {
                    "typst_icon": "passport",
                    "url": None,
                    "clean_url": None,
                    "placeholder": self.nationality,
                }
            )
```

in 

https://github.com/rendercv/rendercv/blob/2adde0b38483202c99f4b1c5214c9a113d4088f7/rendercv/data/models/curriculum_vitae.py#L499

Then, you can just type in your `.yaml `file

```YAML
cv:
  birth: 01.01.2000 in Berlin
  nationality: greek
```
The icons I chose are rendered nicely in the output. 

Thanks again for this great software!

sinaatalay:
This is now possible with `cv.custom_connections` field (see [docs](https://docs.rendercv.com/user_guide/yaml_input_structure/cv/)) in v2.4.

PR #434, opened 2025-06-02
Add 'nationality' and 'birthdate' keys

oiao:
Awesome project! 
I wanted to contribute adding `cv.nationality` and `cv.birthdate` to connections, as discussed in #383  and #407.
Note that the unit tests currently fail, but this seems to be an upstream issue  (e.g. [here](https://github.com/rendercv/rendercv/actions/runs/15381140834/job/43272092002#step:4:648)).

Any additional modifications to the PR are of course welcome


sinaatalay:
I apologize for the late response. I haven’t been maintaining RenderCV for a while, but I’m back now.

I had to close the PR because the tests were failing.

Issue #353, opened 2025-02-11
Dockerfile and documentation improvements


isaacdcm:
First of all, I love this tool you created.  Thanks for your hard work on it.

As a matter of convenience, I really like to run this via the docker image that is created and to keep permissions clean on my linux system, I set the user id and group at the docker commandline.  

I do all of this in the Makefile below which could be useful to add to docs maybe.

```
PWD=$(shell pwd)
CURRENT_UID=$(shell id -u):$(shell id -g) 
IMAGE=rendercv/rendercv:v2.2
DOCKER=docker
DOCKER_RUN=$(DOCKER) run --rm -it  -v $(PWD):/rendercv -u $(CURRENT_UID) -e "XDG_CACHE_HOME=/tmp" $(IMAGE)
RENDERCV=$(DOCKER_RUN) rendercv

render:
	$(DOCKER_RUN) -c "rendercv render ./MY_CV.yaml"

```

I found that when I ran this in version 1, it worked great without that XDG_CACHE_HOME variable, but on version 2 I got the following error

```
An error occurred:

error: failed to decompress package (failed to create `/.cache/typst/packages/preview/fontawesome/0.5.0`)
  ┌─ rendercv_output/MY_CV.typ:2:8
  │
2 │ #import "@preview/fontawesome:0.5.0": fa-icon
  │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The basic issue is that because no home directory exists for the UID I send in, typst tries to write to the root directory `/` and I get an error.

Per the instructions here
https://github.com/typst/packages/blob/main/README.md

Setting that `XDG_CACHE_HOME` variable sets the temp directory, and the base image you are using has a wide open /tmp directory ready to go, so easy fix is add the following to Dockerfile

```
ENV XDG_CACHE_HOME=/tmp
```

The second mildly annoying issue, which maybe is just me being nitpicky, is that I feel like the entrypoint should actually be the rendercv executable so that the command could just be 

```
docker run -it --rm -v ........ rendercv/rendercv render ...
docker run -it --rm -v ........ rendercv/rendercv new ...
```

alternatively, the entrypoint could just be the more traditional approach like below to make this work more like a normal dockerized utility so I dont have to wrap my commands in `-c "......"`.

```
ENTRYPOINT ["/bin/bash", "-c"]
```

sinaatalay:
[...]

PR #435, opened 2025-06-03
Dockerfile changes

net42-jkeil:
This PR changes the `pip install` call so it doesn't keep cache around.
This reduces the image size from 365MB to 307MB (~15.9%)

It also sets `ENTRYPOINT ["rendercv"]` as suggested in #353

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,4 @@ RUN pip install --no-cache-dir "rendercv[full]"
 WORKDIR /rendercv
 
 # Set the entrypoint to /bin/sh instead of Python
-CMD ["rendercv"]
+ENTRYPOINT ["rendercv"]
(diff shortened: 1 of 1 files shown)


net42-jkeil:
Or should we change this to `ENTRYPOINT [""]` to allow seamless CI pipeline integration?

Is there anything you need before you can merge this? @sinaatalay

Dockerfile
```diff
@@ -1,12 +1,11 @@
...
+# Use the official Python image as a base
+FROM python:3.13-slim
+
+# Install RenderCV:
+RUN pip install --no-cache-dir "rendercv[full]"
+
+# Create a directory for the app
+WORKDIR /rendercv
+
+# Set the entrypoint to /bin/sh instead of Python
```
thiagowfx, line 10: comment is outdated

net42-jkeil:
was implemented in #511

PR #441, opened 2025-06-21
Allow user to specify custom columns in a row by specifying || as delimiter

corwinmacmillan:


corwinmacmillan:
Rather than having date_and_location_column, etc., simplify the backend by simply specifying a template for each row. The template will contain || to denote a new column, e.g. "**NAME** -- LOCATION || DATE". This also allows the user to have more freedom over where they want the date inserted.

PR #450, opened 2025-07-07
feat: allow customizing connections order

IanHollow:
* allow changing the order of the connections. Keeps the default order if the new field is not added.


IanHollow:
It might be better to order the connections in the order they are placed in the yaml file. However, that could potentially break old yaml configs looking the same. Honestly though doing it through the order in the yaml makes more sense to me.

PR #451, opened 2025-07-07
add grade field to EducationEntryBase model

IanHollow:
Add a grade field to the education section. This allows the grade to be placed in a more customized way.

### DIFF
--- a/examples/John_Doe_ClassicTheme_CV.yaml
+++ b/examples/John_Doe_ClassicTheme_CV.yaml
@@ -24,6 +24,7 @@ cv:
         summary:
         highlights:
           - Working on the optimization of autonomous vehicles in urban environments
+        grade: "GPA: 3.9/4.0"
       - institution: Boğaziçi University
         area: Computer Engineering
         degree: BS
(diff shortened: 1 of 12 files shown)


IanHollow:
I think that the tests should pass if another PR that i worked on get merged. So i would say that this PR depends On #456.

However I have not had a chance to test this by merging the branches and running the tests locally. If I get a chance to do so I will write back here unless the PR with the fixes gets merged.

IanHollow:
> I think that the tests should pass if another PR that i worked on get merged. So i would say that this PR depends On #456.
> 
> However I have not had a chance to test this by merging the branches and running the tests locally. If I get a chance to do so I will write back here unless the PR with the fixes gets merged.

I tested this with the fixes from PR #456. While the changes from #456 prevent some errors for test cases. This PR seems to cause failing test cases to the rendering test cases. I do not understand how that is the case. As I am just adding an additional field not causing this field to be required or change the examples in any way. There must be some additional aspect that is created causing a secondary effect. @sinaatalay

IanHollow:
Ok it seems that creating a new field in the education entry creates an automatic test variation with that field. I could make a test for this field that uses this but since it is non-standard field it does not appear anyways. I added an ignore option for fields and added this new field to it. All the test cases seem to pass now.

IanHollow:
Fixing the errors seemed to bring a regression to this PR. Now the grade field is not rendering the grade information. That will have to be fixed before this gets merged.

IanHollow:
Superseded by #463

PR #455, opened 2025-07-18
use YAML key order to determine header order

IanHollow:
This another approach to a previous PR #450 that I talked about in the previous approach I personally like this approach much better.

This PR allows for the order of the header fields to determine the order that appear on final output.


sinaatalay:
Wow, thank you, that's great. I'll merge it, but there's a conflict.

PR #456, opened 2025-07-20
fix all test errors and warnings

IanHollow:
I was running into some test errors when creating some other PRs so I decide to fix them.


sinaatalay:
Thanks a lot.

PR #457, opened 2025-07-20
Arbitrary keys bug?

corwinmacmillan:
It seems that this line in templater.py originally would always return None. Since arbitrary keys in the yaml was not working properly, I assume it is meant to return the artbitrary keys. If so, I fixed it, so that it would return arbitrary keys.

### DIFF
--- a/rendercv/renderer/templater.py
+++ b/rendercv/renderer/templater.py
@@ -745,11 +745,13 @@ def replace_placeholders_with_actual_values(
         The string with actual values.
     """
     for placeholder, value in placeholders.items():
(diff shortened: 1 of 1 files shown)


sinaatalay:
Sorry for very late response. Did you test if this fixes the issue? If so, I'll merge.

corwinmacmillan:
Yes, it works in the examples I tested (e.g. arbitrary_key then doing ARBITRARY_KEY in the template section). I can add arbitrary keys to the examples, but I thought this might be cleaner

PR #461, opened 2025-07-21
Create optional automaitc sorting capabilites for entries

IanHollow:
This allows for you to have the ability to sort entries in reverse-chronological, chronological, and no sorting. The default is no sorting.

The date sorting tries to follow the displayed date sorting as closely as possible.


IanHollow:
Additionally there are some test errors but I believe that they should be fixed when PR #456 get merged. Therefore this PR depends on #456.

IanHollow:
@sinaatalay this feature seems to have been removed after your refactor

PR #463, opened 2025-07-22
add grade field to education entry

IanHollow:
Add a grade field to the education section. This allows the grade to be placed in a more customized way.

Supersedes #451.

I rewrote the last PR because that one was getting messy and was not working.

### DIFF
--- a/rendercv/data/models/entry_types.py
+++ b/rendercv/data/models/entry_types.py
@@ -579,6 +579,11 @@ class EducationEntryBase(RenderCVBaseModelWithExtraKeys):
         description="The type of the degree, such as BS, BA, PhD, MS.",
         examples=["BS", "BA", "PhD", "MS"],
     )
+    grade: Optional[str] = pydantic.Field(
+        default=None,
+        title="Grade",
+        examples=["GPA: 3.00/4.00"],
+    )
 
 
 class EducationEntry(EntryBase, EducationEntryBase, EntryType):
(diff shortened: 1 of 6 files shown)


sinaatalay:
I apologize for the very late feedback, but I think RenderCV’s schema should be more rendering-oriented rather than data-oriented. When you include your grade as an item in the highlights, you know exactly how it will be rendered. You can customize it a little with Markdown as well. When it’s a new field, that’s not as clear. The only justification for adding a new field would be if there’s a good idea for customizing the grade display, do you have any ideas on that?

I’m back to maintaining RenderCV now, and it won’t take me months to reply this time. Sorry about that.

IanHollow:
The reason I added this is to allow the ability to place the grade in a specific spot. In my opinion placing the grade as highlight (bullet point) takes unnecessary space on the CV. For example with the grade field you can place your grade (GPA) under the date on the right side of the page. It is not really possible to do this otherwise.

IanHollow:
Here is an example of why I believe this field is needed to be more efficient with the space on the CV:

image

I did not change the examples as I know this is a preference but without these changes doing something like this is not possible.

Prometheus3375:
FYI, this new field is not rendered using classic theme. I even remade classic theme files to ensure it is supported.

IanHollow:
You have to edit the design settings by adding GRADE to get it to work. This was not done by default because it is an optional field.

sinaatalay:
Hey! I reverted this change in #528. Optional fields are now fully supported and tested, so the `grade` field works exactly as you described in this PR without needing to be explicitly defined in the Pydantic models. I'll be releasing v2.4 this week with these changes.

PR #465, opened 2025-08-05
add entry vertical space overrides for specific sections

IanHollow:
If you have a skills section on your resume and use one-line-entry type would use vertical spacing from the entries spacing setting for all entries. However each one line entry might make more sense to have less vertical space as they are just lists of skills. Therefore, I added a way to override the spacing for specific entry types to allow for more customization.


IanHollow:
@sinaatalay Is there anything you would like to see change to merge this into main?

sinaatalay:
I'm not sure if we should merge this. I feel like a one-line entry is just an entry type, and the vertical spacing between them should be the same as other entry types. I tend to keep everything simple unless we really need something.

IanHollow:
@sinaatalay I think I owe you better explanation of what the purpose of this PR is for. The issue I am currently having is that lets say have four oneLineEntry in a section in a row. For me this would be Skill sections something that might look like this:

```yaml
skills:
  - label: Languages
    details: "Python | C | C++ | JavaScript | TypeScript | Swift | Java | SQL | HTML | CSS | Nix"
  - label: Technologies
    details: "Git | Linux | Bash | Docker | Kubernetes | Redis | PostgreSQL | MongoDB | AWS | GCP"
  - label: Libraries
    details: "PyTorch | TensorFlow | Hugging Face Transformers | Pandas | NumPy | Scikit-Learn | OpenCV"
  - label: Frameworks
    details: "React | Next.js | Node.js | Flask | Django"
```

Now without with the current spacing method that is on the main branch each of these oneLineEntry will be spaced by the entries.vertical_space_between_entries as each oneLineEntry is considered a separate entry. Each being treated as entry make sense and I don't think this should change.

However, in this case the skills section that I am trying to create is more supposed to be like one block of information that is a single entry. I do not know a better way of being able to do this with the current structures that are available. What I am trying to achieve is highlights like bullet spacing without the bullet points.

Maybe a better way to implement this functionality to add seperate class to deal with this type of situation. While the use case seems niche it is common for people to add a skills section to the resume/cv and being able to save space is important as well. I would not mind treating each oneLineEntry with entry level spacing but it takes up valuable space on the page.

If you would consider merging something like this with a new class or some other solution let me know I am will be able to implement something like that. In this PR I was trying to create something that was both minimal would not be disruptive as it is a simple override. However things get more complex if you want to treat some of consecutive oneLineEntries with with entry level spacing and other consecutive oneLineEntries with highlights level spacing.

Also, let me know if you have any questions.

sinaatalay:
Would using a TextEntry with line breaks be helpful in this situation?

IanHollow:
@sinaatalay 

Yes, I can write this skills section the way I want like this as you were saying:

```yaml
skills:
  - |
    **Languages:** Python | C | C++ | JavaScript | TypeScript | Swift | Java | SQL | HTML | CSS | Nix

    **Technologies:** Git | Linux | Bash | Docker | Kubernetes | Redis | PostgreSQL | MongoDB | AWS | GCP

    **Libraries:** PyTorch | TensorFlow | Hugging Face Transformers | Pandas | NumPy | Scikit-Learn | OpenCV

    **Frameworks:** React | Next.js | Node.js | Flask | Django
```

However, I feel like this kind of takes away from the whole point as I am basically now writing my skills section in markdown. While I understand that what I want is more custom and more custom problems require more custom solutions. This does seem like a much less elegant solution to the problem though. The final decision for how you want to deal with issue of customization comes down to you because it is your project.

I added an override for spacing for entry types originally as I thought it could be feature outside of just this specific use case but I did not sell this feature well enough in that way. I thought more control and customization would be welcomed if solutions remained simple and not distributive to current designs as this is an optional parameter that would not effect anyone if they did not know it existed.

Anyways, let me know what you think. I am happy to rewrite the code to implement this in a better way as well as I know you working on a big rewrite/refactor.

IanHollow:
Another issue I have with TextEntry is that the line spacing of the TextEntry does not seem possible to set so if I am trying to keep the same spacing as between highlights then it is not necessarily possible unless I am missing something.

sinaatalay:
Yeah, adding an additional design field wouldn’t hurt, I guess. Okay, we can do that. But I’m working on a huge refactoring of RenderCV (#528), where I also will change the Typst code and templating. Would you like to do this PR after that one is ready?

IanHollow:
Sounds good. Thank you for working with me.

sinaatalay:
When I was working on #528 (a massive refactor and improvement), I implemented something that I think addresses your problem:

- `design.sections.space_between_regular_entries`
- `design.sections.space_between_text_entries`

The latter controls the space between entries like `BulletEntry`, `TextEntry`, `NumberedEntry`, and `ReversedNumberedEntry`. Would this be enough to achieve what you want? If not, or if you have other ideas, I'd love to hear them.

I'm closing this PR since it's diverged from main after #528 was merged.

PR #466, opened 2025-08-05
Add more system fonts

IanHollow:
Right now system fonts are not checked to see if they are installed and if the system font is not installed then it fallback to some default font instead of error occurring. 

I think ideally we should check if the font is installed on the system. It is also possible for some of these fonts to be moved to the rendercv-fonts package to prevent having to worry about having to worry about the system having the font installed. However, you would have to deal with the legal implications for some of these fonts.


sinaatalay:
Thanks.

PR #496, opened 2025-09-15
ci(pre-commit): add yamllint

thiagowfx:
Plus fix existing lint violations.

https://github.com/adrienverge/yamllint


.yamllint.yaml
```diff
@@ -0,0 +1,3 @@
+extends: default
+rules:
+  line-length: disable
```
jakob1379, line 3: If you are really nice, you can reduce the amount of littering files if you add this to pyprojec.toml :D

thiagowfx: Do you have an example?

https://github.com/adrienverge/yamllint/issues/568 suggests this is not possible

jakob1379: Wat, by bad then, thought they had closed that long ago!

Usually I use yamlfix over yamllint as it actually does the safe fixes, so I might have mixed them up. Sorry.

sinaatalay:
Thank you.

Issue #503, opened 2025-10-01
Make a new release please


corneliusroemer:
There are lots of issues that were closed when bug fix PRs were merged but they are still not live due to the latest release being >7 months ago. Would be great to have a new release.

TheColorDarkGreen:
bump

sinaatalay:
I apologize. I'm back to maintaining RenderCV. `v2.3` has been released. `v2.4` is coming soon.

PR #509, opened 2025-10-14
ci: trigger publish workflow on semver tags

jakob1379:
## Summary
- automatically run the publish workflow whenever a semver-style tag (including prereleases) is pushed
- guard the pipeline so only properly formatted tags make it through, ensuring releases happen after tests/builds succeed

## Why this helps
This change lets maintainers tag a release and have the PyPI package, binaries, and container images appear without manual intervention, reducing the chance of human error and speeding up the release cycle envisioned in #503.

## Testing
- not run (workflow-only changes)


sinaatalay:
Thank you.

PR #512, opened 2025-10-16
Removed `__pycache__` from the output directory

theobori:
Hello,

This pull request exclude the `__pycache__` Python directory from the output directory. Also see


sinaatalay:
Thank you.

PR #542, opened 2025-12-09
Add i18n support for CV section titles via locale configuration

yacc143:
Enables users to translate section titles by defining a section_titles mapping in the locale configuration. The mapping uses lowercase section keys (as they appear in YAML) to their translated versions.

- Add section_titles field to Locale model with dict[str, str] type
- Implement translate_section_title() function to lookup translations
- Integrate translation into CurriculumVitae.sections property
- Add comprehensive tests for translation and reset behavior

### DIFF
--- a/src/rendercv/data/models/computers.py
+++ b/src/rendercv/data/models/computers.py
@@ -437,3 +437,35 @@ def dictionary_key_to_proper_section_title(key: str) -> str:
         )
         for word in words
     )
+
+
+def translate_section_title(key: str, title: str) -> str:
+    """Translate a section title using the locale section_titles mapping if available.
+    
+    This function takes a section key (as it appears in the YAML file) and the 
+    formatted title, and returns the translated title if a translation is available
+    in the locale configuration.
+    
+    Example:
+        ```python
+        # With locale.section_titles = {"education": "Ausbildung"}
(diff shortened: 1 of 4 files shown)


sinaatalay:
Thanks for the PR! Since section titles in the YAML are already arbitrary, users can write them in any language they want directly. I don't think we need a separate Locale field for this. Am I missing something?

yacc143:
- well an incredible lucky timing, I checked you managed to commit your huge refactor less than two hours earlier. That luck. 

- yes and no, I'm myself not really happy with this, personally I'd love to be able to have one file for keeping my CV in multiple languages (eg field, and field[de] and field[fr] thus having all text but that is a bit hard with the current structure. 

Basically it's me experimenting with my CV and RenderCV and for professional reasons I do need to keep my CV in at least English and German current.

Am 9. Dezember 2025 20:27:32 MEZ schrieb Sina Atalay ***@***.***>:

sinaatalay:
But even if you had locale for section titles, wouldn't you still need to rewrite all the content in German anyway?

jakob1379:
this PR might also fix the issue I had suddenly rendercv yelled that danish is not an available locale, even though I had it defined in the root of my yaml.

#556

Pulastya12 (approved):

Issue #557, opened 2025-12-15
settings.bold_keywords applies to placeholder variables


boxspah:
Here is the relevant part of my yaml file:
```yaml
settings:
  current_date: '2025-12-15'
  bold_keywords:
    - John Doe  # same as cv.name
  render_command:
    pdf_path: NAME_IN_SNAKE_CASE_CV.pdf
```

**Expected:** The output PDF is named `John_Doe_CV.pdf`.
**Actual:** The output PDF is named `#strong[John_Doe]_CV.pdf`.

**To help with debugging:** When I set `pdf_path` explicitly to `John_Doe_CV.pdf`, the output filename is `John_Doe_CV.pdf`. I'm using v2.5.

mbaprogress008-stack:
"🚨 Flame Guardian 🔥💧
Hi, devs! 👋
I'm working on Water Bomb Safety Device - a smart fire safety project detecting fires & sprinkling water to quench flames 🚀
Need your feedback/collaboration! 😊
Project
Let's make the world safer! 🌟"

sinaatalay:
I merged it but I think #575 doesn't address this issue.

sinaatalay:
Fixed in v2.7. `bold_keywords` no longer affects file path placeholder variables like `NAME_IN_SNAKE_CASE`.

PR #575, opened 2025-12-24
fix the filename in pdf header rendercv/rendercv#557

abidkhan484:
**Issue**: rendercv/rendercv#557

The case has been happened due to the usage of `name` in the rendercv_model instead of using `plain_name`. The fixation is done in the `src/rendercv/renderer/templater/templates/typst/Preamble.j2.typ` file by using the right attribute.


abidkhan484:
@sinaatalay, could you please take a look at the changes?

sinaatalay:
Thank you.

PR #577, opened 2025-12-24
fix: allow PDF/PNG generation when dont_generate_typst is True

### LINKED ISSUE #550: [Bug] Disabling typst generation disables overall rendering
For my purposes, I just wanted direct PDF, no other files.
So I had configured as such:
```yaml
settings:
  current_date: "2025-12-14"
  render_command:
    design:
    locale:
    html_path: rendercv_output/NAME_IN_SNAKE_CASE_CV.html
    markdown_path: rendercv_output/NAME_IN_SNAKE_CASE_CV.md
    png_path: rendercv_output/NAME_IN_SNAKE_CASE_CV.png
    typst_path: rendercv_output/NAME_IN_SNAKE_CASE_CV.typ
    pdf_path: rendercv_output/NAME_IN_SNAKE_CASE_CV.pdf
    dont_generate_markdown: true
    dont_generate_html: true
    dont_generate_typst: true
    dont_generate_pdf: false
    dont_generate_png: true
```
I'm using "rendercv[full]>=2.5"

I looked into readme, but there was nothing I could find relating to my issue. 
So I created a copy of my CV and dumped lorem ipsum, turns out there was no validation issue either.

Turns out disabling typst, disables entire file generation.

`tempfile` library can be used to set typst path if it is disabled in yaml config.

majiayu000:
## Summary
- Fixed issue where setting `dont_generate_typst: true` would also disable PDF and PNG generation
- Now uses a temporary file for typst compilation when the user doesn't want to save the .typ file

## Changes
- Modified `generate_typst()` to return a temp file path instead of None when `dont_generate_typst=True`
- Updated `generate_pdf()` and `generate_png()` to always expect a valid typst path
- Added regression tests to verify PDF/PNG generation works with `dont_generate_typst=True`
- Updated existing test expectations to reflect the new behavior

Fixes #550

### DIFF
--- a/src/rendercv/cli/render_command/run_rendercv.py
+++ b/src/rendercv/cli/render_command/run_rendercv.py
@@ -1,4 +1,5 @@
 import pathlib
+import shutil
 import time
 from collections.abc import Callable
 from typing import Unpack
@@ -97,6 +98,7 @@ def run_rendercv(
         progress: Progress panel for output display.
         kwargs: Optional overrides for design/locale files, output paths, and generation flags.
     """
+    temp_typst_dir: pathlib.Path | None = None
     try:
         _, rendercv_model = timed_step(
             "Validated the input file",
@@ -105,12 +107,14 @@ def run_rendercv(
             main_input_file_path_or_contents,
             **kwargs,
         )
-        typst_path = timed_step(
+        typst_path, is_temp_typst = timed_step(
             "Generated Typst",
             progress,
             generate_typst,
             rendercv_model,
         )
+        if is_temp_typst:
+            temp_typst_dir = typst_path.parent
         timed_step(
             "Generated PDF",
             progress,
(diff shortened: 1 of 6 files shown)


sinaatalay:
Are we sure those temp dirs get removed at the end of the run?

majiayu000:
Good catch! You're absolutely right - the temp directories were not being cleaned up.

I've pushed a fix that:

1. **Modified `generate_typst()`** to return a tuple `(path, is_temporary)` so the caller knows when cleanup is needed
2. **Added cleanup logic in `run_rendercv()`** using a `finally` block to ensure temp directories are removed even if an error occurs during PDF/PNG generation
3. **Added tests** to verify the cleanup behavior

The temp directory is now properly cleaned up after the PDF/PNG compilation is complete. All 444 tests pass.

Thanks for the review!

MRDGH2821:
OS's temp dir could also be used.
In case of windows, manual cleanup would be still required or the user needs to use Disk Cleaner to get it removed.

PR #579, opened 2025-12-24
feat: watch mode now monitors included config files

### LINKED ISSUE #552: Watch mode should watch for changes in included files
Watch mode is great as it is, since it allows to quickly verify how a specific change impacts the overall look of the CV, especially when tinkering with the **design** options.

However, when using the `--design` parameter to separate the design configuration from the actual CV data, the **watch mode does not listen for changes in that file** as one would have expected.

This also applies to all the other file parameters (`--locale-catalog` and `--settings`).

It goes without saying that having the watch mode disabled when it is needed the most (_eg. playing around with fonts or colours to find what's best_) kinda defies its purpose 😄

majiayu000:
## Summary
- Watch mode now monitors included configuration files (--design, --locale-catalog, --settings)
- Changes to any watched file trigger a re-render

## Changes
- Modified `run_function_if_file_changes()` to accept multiple file paths
- Updated `cli_command_render()` to pass all config files to the watcher
- Added test to verify config files are included in watch list

## Example
```bash
rendercv render cv.yaml --design design.yaml --watch
# Now changes to either cv.yaml OR design.yaml will trigger re-render
```

Fixes #552


sinaatalay:
Thank you!

Issue #452, opened 2025-07-09
How do i manipulate this to display arabic


vigneshmj1997:
I have reviewed the documentation and noticed that there is no mention of support for the Arabic format. Specifically, I tried setting data_model.locale.language = "ar", but this did not adjust the heading or align the skill section correctly from right to left.

It would be helpful to have documentation or guidance on enabling proper right-to-left (RTL) support for Arabic, including heading alignment and layout adjustments.

sinaatalay:
We would need to play with Typst for that. If someone is willing to help on this, that would be great. Please see the [developer guide](https://docs.rendercv.com/developer_guide/).

sinaatalay:
Fixed in v2.7. Arabic is now supported as a built-in RTL locale.

PR #587, opened 2025-12-24
feat(locale): add Hebrew locale (and generic RTL issues)

DeepSpace2:
(Related to #452 and blocked by #591)

Added Hebrew as a valid locale. Also took the liberty to update line numbers in the docs.

There are still quite a few RTL issues.

PDF:
 - Sections' headers are rendered correctly, however the contents is still aligned to the left:
 
image

<hr>

HTML:
 - The rendered html is missing `<body dir="rtl">`. Can be manually added by the user but would be nice if could be done automatically. The fix for this might required adding a new "rtl" boolean to locale in the schema/model.

<hr>

There are possibly more issues that I haven't found or noticed yet.

As the fixes would probably be required to render other RTL languages we can use this PR as a larger initiative.

This feels like a regression from 2.3, as the CV is rendered perfectly on the web app by simply using

```yaml
locale:
  language: he
```

image

### DIFF
--- a/docs/developer_guide/how_to/add_locale.md
+++ b/docs/developer_guide/how_to/add_locale.md
@@ -48,7 +48,7 @@
 
 3. Add ISO 639-1 language code to `english_locale.py`
 
-    Edit `src/rendercv/schema/models/locale/english_locale.py` line 95-108:
+    Edit `src/rendercv/schema/models/locale/english_locale.py` line 97-114:
 
     ```python
     return {
--- a/schema.json
+++ b/schema.json
@@ -801,6 +801,100 @@
       "title": "GermanLocale",
       "type": "object"
     },
+    "HebrewLocale": {
+      "additionalProperties": false,
+      "properties": {
+        "language": {
+          "const": "hebrew",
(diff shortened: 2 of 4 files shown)


sinaatalay:
Thank you! See #591 and https://github.com/rendercv/rendercv-typst/pull/3

DeepSpace2:
Nice, so apparently the PDF RTL issue is solved by #591 and https://github.com/rendercv/rendercv-typst/pull/3.  I can wait for these to be merged and verify against my branch.

As for the HTML, I'll need to look a bit deeper as it doesn't look like it is handled by #591. Again, I'll wait for it to be merged as I need access to the added properties in https://github.com/rendercv/rendercv/blob/main/src/rendercv/renderer/html.py

anefzaoui:
Hi @sinaatalay 
Any plan to get this merged soon?

sinaatalay:
> Hi @sinaatalay Any plan to get this merged soon?

Yes @anefzaoui , support for right-to-left languages is coming very soon in v2.7.

PR #589, opened 2025-12-25
replace asserts with explicit error handling

kellemar:
asserts in production code get stripped out when python runs with -O flag, which could cause weird silent failures

swapped them out for explicit if checks that raise RenderCVInternalError instead. same behavior, but now it actually works regardless of how python is invoked

touched 7 files, 15 asserts total - all tests still pass


sinaatalay:
This is perfect, thank you!

PR #597, opened 2025-12-25
Fix: Prevent local file access via YAML path fields (path containment) - Closed due to inefficient fix

alpernae:
This PR fixes a local file access vulnerability where YAML path fields (e.g., `cv.photo`) could reference arbitrary absolute paths outside the input directory, leading to potential disclosure of sensitive files during rendering.

**Key changes:**
- The path resolution logic in `resolve_relative_path` now enforces that all resolved paths must be contained within the input YAML's directory. Absolute or escaping paths are rejected.
- Added a test (`test_yaml_path_containment.py`) to verify that paths outside the input directory are not allowed, and that valid relative paths within the directory are accepted.

**Impact:**
- This closes the vulnerability where a crafted YAML could cause the renderer to access or copy arbitrary files from the host system.
- The fix is enforced at the schema validation layer, so all downstream consumers are protected.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+log_cli_level = INFO
+xfail_strict = true
+addopts = -ra -v --strict-markers --strict-config
+testpaths = tests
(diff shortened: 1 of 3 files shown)


alpernae:
**Closed due to inefficient fix**

Issue #598, opened 2025-12-25
Security Issue: Local file access / Path exposure via YAML file paths [Medium]


alpernae:
# Local file access / Path exposure via YAML file paths

## Description

I identified a local file access vulnerability in `rendercv` where user-supplied YAML file path fields (for example `cv.photo`) can reference arbitrary absolute paths on the host filesystem. The current path resolution logic verifies file existence but does not enforce that resolved paths remain within the intended input directory. As a result, a crafted YAML file can cause the renderer to read or copy sensitive local files during document generation.

This issue can lead to unintended disclosure of host files if generated artifacts or render directories are accessible to untrusted users (for example in web services, CI pipelines, or shared environments).

---

## Technical details

The vulnerability originates from how file paths are validated and later consumed:

* YAML path fields such as `cv.photo` are validated using the `ExistingPathRelativeToInput` logic.
* The validator calls `resolve_relative_path()` in `src/rendercv/schema/models/path.py`.
* `resolve_relative_path()`:

  * Accepts absolute paths without restriction.
  * Only checks that the path exists when `must_exist` is set.
  * Does not enforce that the resolved path is contained within the input YAML’s directory.
* Later, rendering code uses the resolved path directly as a file source. For example, in `src/rendercv/renderer/pdf_png.py`, the renderer copies the referenced file using `shutil.copy(...)`.

Because no containment check is performed, absolute paths such as `/etc/hosts` or `/home/user/.ssh/id_rsa` are treated as valid inputs and are accessed during rendering.

---

## Steps to reproduce

1. Create a YAML file that references an absolute path outside the input directory:

```yaml
cv:
  name: Reader
  photo: /etc/hosts
```

2. Run the renderer with this YAML file:

```bash
rendercv render cv.yaml --pdf-path output.pdf
```

3. Observe that during generation the renderer attempts to access and copy `/etc/hosts` into the Typst/render workspace.

**Expected behavior:**
The resolver should reject absolute paths or paths that escape the input directory.

**Observed behavior:**
The renderer accepts the absolute path and accesses the referenced file.

---

## Proof of Concept

The following commands demonstrate the issue using a benign system file:

```bash
# Create YAML input
cat > cv.yaml <<'YAML'
cv:
  name: Reader
  photo: /etc/hosts
YAML

# Run the renderer
rendercv render cv.yaml --pdf-path output.pdf
```

[...]

PR #601, opened 2025-12-25
Fix: Prevent local file access via `photo` field using magic bytes

alpernae:
# Summary

This PR addresses the local file access vulnerability by implementing file type validation for the `cv.photo` field. Instead of restricting paths ([597](https://github.com/rendercv/rendercv/pull/597)) to the input directory (which limits user flexibility), this approach validates that the referenced file is actually a valid image by checking its magic bytes.

Changes
`src/rendercv/schema/models/path.py`

Added:

+ `IMAGE_SIGNATURES` constant: Magic byte signatures for `PNG, JPEG, GIF, BMP, WebP`, and `TIFF` formats.

+ `is_valid_image_file(path)` function: Validates a file is an image by reading and checking its first 12 bytes against known signatures.

+ `validate_image_path(path, info)` function: Combines path resolution with image format validation.

+ `ExistingImagePathRelativeToInput` type alias: New Pydantic-annotated type for image-specific path validation.

src/rendercv/schema/models/cv/cv.py
**Modified**:
+ Changed `photo` field type from `ExistingPathRelativeToInput` to `ExistingImagePathRelativeToInput`
+ Updated field description to document supported image formats.

## Related Issues

+ Fixes > https://github.com/rendercv/rendercv/issues/598

### DIFF
--- a/src/rendercv/schema/models/cv/cv.py
+++ b/src/rendercv/schema/models/cv/cv.py
@@ -7,7 +7,7 @@ import pydantic_extra_types.phone_numbers as pydantic_phone_numbers
 from rendercv.exception import RenderCVInternalError
 
 from ..base import BaseModelWithExtraKeys
-from ..path import ExistingPathRelativeToInput
+from ..path import ExistingImagePathRelativeToInput
 from .custom_connection import CustomConnection
 from .section import BaseRenderCVSection, Section, get_rendercv_sections
 from .social_network import SocialNetwork
@@ -41,9 +41,9 @@ class Cv(BaseModelWithExtraKeys):
             ["john.doe.1@example.com", "john.doe.2@example.com"],
         ],
     )
-    photo: ExistingPathRelativeToInput | None = pydantic.Field(
+    photo: ExistingImagePathRelativeToInput | None = pydantic.Field(
         default=None,
-        description="Photo file path, relative to the YAML file.",
+        description="Photo file path, relative to the YAML file. Must be a valid image file (PNG, JPEG, GIF, BMP, WebP, TIFF).",
         examples=["photo.jpg", "images/profile.png"],
     )
     phone: (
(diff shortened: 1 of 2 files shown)


LukeSavefrogs:
This is already better but I would think about using an already well-tested and known package for the role of detecting the actual mime type instead of reinventing the weel...

After a quick search I found the following to be the most appetizing options (always imo):
- [Pillow](https://stackoverflow.com/a/53470882/8965861): probably overkill, since we would only use the `verify` method (the `transpose` too if we want to be fancy)
- [filetype](https://stackoverflow.com/a/61195193/8965861): also uses magic bytes to detect the actual file type but is more structured and allows passing a stream (such as an online resource) since it only needs the first bytes anyway.

These were just 2 cents, since I hate having to maintain something that someone already worked hard to do and it is for sure more tested and used than my version... 😄 

@sinaatalay what do you think?

It is for sure lighter to just add the code rather than using an external package, but maybe for the future it is better to use a package and just focus on using the result instead of the implementation.

alpernae:
Hi @sinaatalay,

How would you like to proceed here?

LukeSavefrogs:
@alpernae in the meantime can you pls fix the failing test? 

This way @sinaatalay would have everything ready

alpernae:
@LukeSavefrogs All issues should be fixed now.

sinaatalay:
Thank you. Could you provide an example of how a bad actor might exploit this vulnerability, and what the consequences would be?

alpernae:
Hi @sinaatalay,

A security policy with a supported versions table helps users clearly understand which releases address known vulnerabilities (such as the local file exposure issue) and when they should upgrade. This is especially important for a CLI tool, where users may not update automatically.

From a project perspective, documenting your security process demonstrates a good-faith effort to handle vulnerabilities responsibly. It can also reduce legal risk and minimize noise from duplicate or scattered reports. A clear policy directs researchers to the appropriate reporting channel instead of having security concerns posted publicly.

Most importantly, when users evaluate RenderCV, seeing an active security policy signals competence and care—particularly when there is a known file exposure issue on record. The policy itself doesn’t fix the vulnerability, but it gives you control over how issues are reported, handled, and communicated.

> considering that RenderCV is a CLI application without network access?

Being a CLI application doesn’t eliminate risk. Since the project is open source, anyone can download and use it publicly or integrate it into internal services (for example, allowing users to generate CVs). In such scenarios, if user input is not properly validated, a malicious actor could supply a crafted file path. If they also have access to directories containing generated CVs or related files, this could potentially lead to unintended file exposure.

Issue #608, opened 2025-12-26
Quiet CLI option not really quiet


alefteris:
When I tried using the `--quiet` option, I noticed that it is still printing something:

```console
$ rendercv render cv.yaml --quiet
╭─ Rendering your CV... ──────────────────────╮
│ ...                                         │                                                                                                                                                                                                             
╰─────────────────────────────────────────────╯
$
```

It still printing a rich box and an ellipsis. Using rendercv 2.6.

sinaatalay:
Fixed in v2.7. The `--quiet` flag now produces zero output.

PR #610, opened 2025-12-26
fix render --quiet option issue #608

pramanandasarkar02:
I have solved the issue #608 render --quiet option. Please review it


sinaatalay:
Thank you.

Issue #595, opened 2025-12-25
Overriding a YAML value doesn't work when using separate file for design or locale


alefteris:
Overriding a YAML value from the CLI render sub-command doesn't work when using separate files for the design or locale YAML. For example:

```console
$ rendercv render cv.yaml --design design.yaml --design.theme "moderncv"
KeyError: 'design'
```

camggould:
Design fields cannot be overridden via CLI, you should instead update your design file.

rendercv override arguments are designed to override fields in the `cv` model (this would be your cv YAML). The reference to the `design` yaml file is stored in the `settings` portion, separate from the `cv`, as a `PosixPath`, and is not parsed until after the overrides are applied.

[This code path](https://github.com/rendercv/rendercv/blob/main/src/rendercv/schema/rendercv_model_builder.py#L174) reads the CV and applies the overrides passed in. It also persists a `settings` object that contains a reference to the file path of the design yaml, if provided. Afterwards, [this code path](https://github.com/rendercv/rendercv/blob/main/src/rendercv/schema/rendercv_model_builder.py#L180) will read the design file path and validate the contents before adding to the model.

Note that cv value overrides still work fine when including a separate design file:

```
rendercv render camgould_cv.yaml --design design.yaml --cv.name "Test update name"
```

alefteris:
Thanks @camggould for investigating 🙇 .

If design fields are not meant to be overridden via the CLI, the following line should be removed from the CLI reference docs:

https://github.com/rendercv/rendercv/blob/8cac0e9984f426a83ba4ad2a5bcdf48012f08e3e/docs/user_guide/cli_reference.md?plain=1#L145

BTW as I also had noticed, from the exception stack trace and @camggould explained above, that the settings fields can be overridden. For example this works and applies the design settings:

```console
rendercv render cv.yaml --settings settings.yaml --settings.render_command.design 'design-other.yaml'
```

I don't think this is useful for anyone though, since the user can just use the CLI flags to provide the render sub-command options. Except from the `current_date` that doesn't have an equivalent CLI flag.

camggould:
I also messed with the `--settings.render_command.design` change too, agreed it is not particularly useful. The logic actually just looks through the entire input dict, which includes both the `cv` key (including all the details of the cv) and the `settings` key (including references to specified files for additional settings).

I also agree that the ability to override the current date may be useful in some contexts.

[...]

PR #612, opened 2025-12-26
Fix: Update user guide CLI reference to remove incorrect override option #595

camggould:
The existing documentation for field overrides is incorrect, and contains a command that claims the user is able to use dot notation to override YAML values in the `design` which is not supported:

> `rendercv render CV.yaml --design.theme "moderncv"`

(See "Override any YAML value" on [this page](https://docs.rendercv.com/user_guide/cli_reference/))

The new documentation was added in commit [33d15d0](https://github.com/rendercv/rendercv/commit/33d15d0f910f9688bd20ca58356e46fcd2f6cf18#diff-7beaa42f1809eaf10e957c3502bcc6be8e5689b17b07eb035006a0ea8ca80f18).

Only fields that exist within `CV.yaml` or `settings.yaml` can be overridden using dot notation, as the [method responsible for handling these overrides](https://github.com/rendercv/rendercv/blob/main/src/rendercv/schema/rendercv_model_builder.py#L33) is not reading the contents of the `design.yaml` file, and only has access to the file location reference in the settings.

This commit updates the documentation to reflect this behavior.

I tested the `rendercv render CV.yaml --design.theme "moderncv"` command to confirm it was unsupported and also dug through the code to localize the issue while investigating #595.

Documentation validated visually via `just serve-docs`.

### DIFF
--- a/docs/user_guide/cli_reference.md
+++ b/docs/user_guide/cli_reference.md
@@ -137,12 +137,12 @@ rendercv render John_Doe_CV.yaml --pdf-path ~/Desktop/MyCV.pdf
 
 **Override any YAML value:**
 
-Use dot notation to change specific fields. This overrides values in the YAML without editing the file.
+Use dot notation to change specific fields. This overrides values in the YAML without editing the file. Overrides are supported for fields within the `CV.yaml` and `settings.yaml` files.
 
 ```bash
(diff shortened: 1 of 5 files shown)


sinaatalay:
Thank you. I’m closing this because we’re supposed to support those overrides as well; it’s probably a bug that we need to fix.

PR #614, opened 2025-12-27
updated clean_url function to properly remove "www" from the url

kellemar:
clean_url() docstring said it strips www. from urls, but the implementation didn't actually do that.

Added the missing .replace("www.", "") call so the function matches what it claims to do. 

Updated the test expectation too since it was checking for the old broken behavior.


sinaatalay:
Thank you! Maybe we should just keep `www`, some people may want to provide it. I'll just update the docstrings instead.

PR #626, opened 2026-01-03
feat: Added feature to populate yaml from existing CV (docx only)

imwaqas:


sinaatalay:
Thank you!

I think simply using ChatGPT, along with a lot of copy-pasting from the RenderCV documentation and your previous CVs, is the way to go for this problem.

PR #629, opened 2026-01-07
Update docker run command in docs

TBG-FR:
Updated [Docker command shown in Getting Started](https://docs.rendercv.com/user_guide/#__tabbed_1_4) docs
- Added `-u` to avoid permissions issue (rendercv has UID `999`, usually a linux user has UID `1000`)
- Added `--rm` to avoid creating leftover containers

**Fixes `PermissionError: [Errno 13] Permission denied: 'Your_Name_CV.yaml'` when running the Docker command from docs**

### DIFF
--- a/docs/developer_guide/dockerfile.md
+++ b/docs/developer_guide/dockerfile.md
@@ -22,7 +22,7 @@ But Docker makes sense if you want:
 
 The RenderCV Docker image is a ready-made environment with Python and RenderCV pre-installed. Just run:
 ```bash
-docker run -v "$PWD":/work -w /work ghcr.io/rendercv/rendercv new "Your Name"
+docker run --rm -v "$PWD":/work -u $(id -u):$(id -g) -e HOME=/tmp -w /work ghcr.io/rendercv/rendercv new "Your Name"
 ```
 
 ## How the Image Gets Published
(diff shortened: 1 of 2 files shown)


TBG-FR:
There's still an issue while trying to render the CV (after `.typ` generation, during `.pdf` generation), probably caused by the usage of a non-root / non-rendercv user 🤔 

```
╭─ Rendering your CV... ───────────────────────────────────────────────────────╮
│ ✓ 252 ms   Generated Typst:                                                  │
│ ./rendercv_output/Test_CV.typ                                                │
╰──────────────────────────────────────────────────────────────────────────────╯
[...]
TypstError: failed to load package (failed to create temporary package directory: Permission denied (os error 13))`

sinaatalay:
Thanks for this fix! I pushed a couple of additions on top of your change:

- **Added `-e HOME=/tmp`**: Without this, `rendercv render` fails because Typst needs a writable home directory to cache packages. With `-u`, the arbitrary UID has no home inside the container, so Typst gets `Permission denied` when creating its temp directory. I verified this by testing both `new` and `render` with Docker.
- **Updated `docs/developer_guide/dockerfile.md`** with the same command for consistency.
- Squashed and rebased on main.

PR #631, opened 2026-01-08
Add user guide for gitea actions workflow creation

Beariful:
Added: docs/user_guide/how_to/set_up_gitea_action_for_rendercv.md

Simple guide for using a gitea action workflow to run RenderCV on a repository containing your CV.yaml file.


sinaatalay:
Thank you. I'm not a Gitea user so I cannot verify this. So I'll close it.

PR #632, opened 2026-01-08
feat: skillicons integration

Instelce:
- SkillIconsEntry to display skill icons from [skillicons.dev](https://github.com/tandpfun/skill-icons).
- `skillicons` attribute to `ExperienceEntry` and `NormalEntry`.
- Documentation and an example

## Preview
Preview

## How to use

```yaml
- icons: py,cpp,rust,go
  size: 20
  theme: dark
```


sinaatalay:
Thank you, but we try to keep RenderCV somewhat opinionated about how resumes should be written. For example, we do not support two-column CVs. We also would not add icons by default; however, users can include them by using custom Typst commands and templates. There are many discussions online about why icons are not preferred in CVs.

PR #642, opened 2026-01-16
fix: :bug: design file not applying with settings file

### LINKED ISSUE #640: `--settings` flag overwrites `--design` path, causing custom design to be ignored
## Description:
  When using both `--settings settings.yaml` and `--design design.yaml` together, the custom design is silently ignored and the CV renders with default design instead.

  ## Steps to Reproduce
  1. Create a design file with custom settings (e.g., custom font)
  2. Create a settings file with output paths
  3. Run: `rendercv render cv.yaml --settings settings.yaml --design design.yaml`
  4. Check the generated CV - it uses default design instead of the custom one

  ## Expected Behavior
  Both settings (output paths) and design (custom theme/font) should be applied together.

  ## Actual Behavior
  Only settings are applied; design file is completely ignored (no error message).

  ## Root Cause
  In `rendercv_model_builder.py`, the yaml_overlays loop processes keys in order: design → locale → settings. When settings is processed, it replaces the entire `input_dict["settings"]` dict, deleting the design path that was set earlier.

tanmay-pathak:
## Summary
  - Closes #640 
  - Reorder overlay processing so later overlays can override earlier ones consistently
  - Ensure `settings.render_command` is initialized after applying overlays to avoid missing-dict errors

## Testing
- Added 2 new tests
- All tests pass

### DIFF
--- a/src/rendercv/schema/rendercv_model_builder.py
+++ b/src/rendercv/schema/rendercv_model_builder.py
@@ -61,11 +61,11 @@ def build_rendercv_dictionary(
     input_dict = read_yaml(main_input_file_path_or_contents)
     input_dict.setdefault("settings", {}).setdefault("render_command", {})
 
-    # Optional YAML overlays
+    # Optional YAML overlays (process settings first to avoid overwriting design/locale paths)
     yaml_overlays: dict[str, pathlib.Path | str | None] = {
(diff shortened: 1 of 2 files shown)


sinaatalay:
Thanks a lot.

PR #652, opened 2026-01-27
feat(locale): Added Norwegian (Bokmål and Nynorsk) locale

andersgaustad:
Added the two official writing standards for Norwegian: [Norwegian Bokmål](https://en.wikipedia.org/wiki/Bokm%C3%A5l) and [Norwegian Nynorsk](https://en.wikipedia.org/wiki/Nynorsk).

Updated `english_locale-py` with the ISO 639-1 codes of 'nb' (for Bokmål) and 'nn' (for Nynorsk).


sinaatalay:
Thank you!

I pushed a small fixup commit with the following corrections: the Nynorsk `present` translation was misspelled as "noverande." The standard Nynorsk form is "nåverande" (with å), according to Claude. Both YAML files were also missing a trailing newline. Finally, `schema.json` should be generated via `scripts/update_schema.py` rather than being manually edited, which is how this project keeps the schema in sync with the code.

andersgaustad:
Nice 👍

> Finally, `schema.json` should be generated via `scripts/update_schema.py` rather than being manually edited

Ah, that makes sense. I think I just used another PR as an example, and saw that `schema.json` was changed there. 

 Thanks for the fixes!

PR #654, opened 2026-01-30
CV filtering with tags

### LINKED ISSUE #164: Filter on content or sections

akib1689:
### Overview

This PR introduces a tag based CV filtering system that allows you to generate different versions of you CV from a single YAML file. I came up with this solution to solve a specific need of mine: tailor different CV for specific audiences (like academic, backend, devops) without maintaining multiple files or worse! multiple git branches.

#### Tags on entry types

- Every CV entry now supports an optional `tags` field.
- New `TextEntry` type for tagged text entries.

#### Tags on highlights

Individual entries in the highlights can now have tags allowing granular control.

```yaml
highlights:
  - "Always included highlight"  # No tags = always included
  - content: "Academic-focused achievement"
    tags: [academic]
  - content: "Industry-focused achievement"
    tags: [industry]
```

#### Version definitions

Now, you can define versions with `include` and/or `exclude` tag filters

```yaml
versions:
  - name: academic
    include: [research, publications, teaching]
  - name: industry
    exclude: [academic]
  - name: devops
    include: [infrastructure, cloud]
    exclude: [deprecated]
```

#### Permissive filtering

- Entries or highlights without tags are always included
- Only entries with tags are subjected to filtering
- Empty sections are always excluded from output.

### How to use

- Step 1: Add tags to your entries

```yaml
cv:
  sections:
    experience:
      - tags: [industry, software]
        company: TechCorp
        position: Senior Engineer
        highlights:
          - "Led team of 10 engineers"
          - content: "Published internal white paper"
            tags: [academic]
      
      - tags: [academic, research]
        company: University Lab
        position: Research Assistant
        highlights:
          - "Published 3 papers in top venues"
    
    publications:
      - tags: [academic]
        title: "Deep Learning for XYZ"
        authors: ["John Doe"]
        journal: NeurIPS 2023
```
- Step 2: define versions

```yaml
versions:
  - name: academic
    include: [academic, research, publications]
  
  - name: industry
    exclude: [academic]
  
  - name: ml-focused
    include: [machine-learning, deep-learning]
    exclude: [outdated]
```

- Generate specific version

```sh
# Generate the academic version
rendercv render my_cv.yaml --version academic

# Generate the industry version
rendercv render my_cv.yaml -V industry

# Generate without filtering (all entries)
rendercv render my_cv.yaml
```

### Filtering logic

- Include Filter: Entry must have at least one of the specified tags
- Exclude Filter: Entry must NOT have any of the excluded tags
- Combined: If both are specified, include is applied first, then exclude
- Permissive: Entries without tags pass all filters automatically

### Backward Compatibility
This feature is fully backward compatible:

- Existing YAML files work without modifications
- Entries without `tags` are always included
- The `--version` flag is optional

Fixes: #164


akib1689:
@sinaatalay Please share your feed back regarding this implementation.

sinaatalay:
Thank you for your contribution. I will review it as soon as possible.

sinaatalay:
Thank you, but I think this feature makes the schema too complicated. While it adds the functionality, it also makes the schema and editor suggestions much more complex, even for users who won’t use this feature.

Also, what do you think about having AI handle all of this for you in a new file with a single prompt?

If we want this feature, we need to come up with a strong UX design that doesn’t make the schema significantly more complicated. This approach also introduces rendering concerns into the schema and validation. Currently, the schema doesn’t consider the rendering side at all.

akib1689:
I think we should gather more feedback. To me this feature should follow opt-out approach. the default schema should suggest the existing flow. Everything in this feature can be made optional.

heavyimage:
FWIW I love this idea.

PR #667, opened 2026-02-12
fix: prevent degree title clipping and hyphenation in classic theme

sandokanCat:
Description
This PR fixes a layout issue in the classic theme where degree titles are often clipped or hyphenated incorrectly due to a very narrow hardcoded column width.

The Problem
In the current implementation, the degree-column in the Typst template has a fixed width of 1cm. While this might suffice for short acronyms like "GPA", it is insufficient for many professional degree titles, especially in languages like Spanish or Catalan, where titles like "Certificado Profesional" or "Máster" are common. This leads to:

Aggressive and often incorrect hyphenation (e.g., "certifica-do").
Text clipping or overlapping when hyphenation is not possible.
Poor visual separation between the degree title and the institution name.
Proposed Changes
Increased Column Width: Adjusted the degree-column-width from 1cm to 2.4cm in the Typst preamble to accommodate longer titles comfortably.
Disabled Hyphenation: Added #set text(hyphenate: false) specifically for the degree column to ensure titles remain intact and readable.
Improved Gutter Spacing: Added an additional 0.4cm to the column-gutter between the degree and the main column to improve legibility and provide better visual breathing room.
Template Override: The changes were implemented as a clean override in the education-entry function within the Typst rendering logic.
Impact
Users with non-English CVs or those with long degree titles will see a significant improvement in the professional appearance of their education section, with no more broken words or cramped layouts.

### DIFF
--- a/src/rendercv/renderer/templater/templates/typst/Preamble.j2.typ
+++ b/src/rendercv/renderer/templater/templates/typst/Preamble.j2.typ
@@ -81,3 +81,48 @@
     day: {{ settings.current_date.day }},
   ),
 )
+
+// Override the education-entry function to increase the degree column width:
+#let education-entry(main-column, date-and-location-column, degree-column: none, main-column-second-row: none) = {
+  metadata("skip-content-area")
+
+  context {
+    let config = rendercv-config.get()
+    let entries-space-between-columns = config.at("entries-space-between-columns")
+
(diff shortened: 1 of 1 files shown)


sinaatalay:
Thank you. This PR should go to the Typst package of RenderCV: https://github.com/rendercv/rendercv-typst

PR #683, opened 2026-03-01
feat(locale): add Polish support and complex pluralization

juliarzymowska:
## Description
This PR introduces support for the Polish locale and refactors the duration calculation logic to support languages with complex pluralization rules.

## The Problem

Previously, RenderCV utilized a binary pluralization logic (singular `if n=1, else` plural). This is insufficient for languages like Polish, where nouns decline into three or more forms depending on the numeral:

- One (Singular): 1 miesiąc
- Few (Paucal): 2 miesiące, 22 miesiące (ends in 2, 3, 4; except 12, 13, 14) 
- Many (Genitive Plural): 5 miesięcy, 12 miesięcy 

Using a simple string for the months or years field forced users to choose one incorrect form for durations like "5 months" or "22 years".

## Key Changes
1. Updated EnglishLocale to support both str (for backward compatibility) and dict[str, str] for complex mappings.
2. Created plural_rules.py to house the mathematical logic for selecting categories like one, few, and many.
3. Refactored `compute_time_span_string` in date.py to use a universal helper function, `_get_localized_label`. This ensures that years and months follow the exact same linguistic rules.
4. Added polish.yaml with grammatically accurate forms for all time-span scenarios.

## How to add new complex language
After following instructions to add a new language new contributors should:
### Identify plural rules
Check the https://www.unicode.org/cldr/charts/48/supplemental/language_plural_rules.html for target language (link is also in `plural_rules.py`).

### Implement logic in `plural_rules.py`
Add language's numeric logic to `plural_rules.py`:
- Define a rule function: This function should take an integer n and return a string matching a CLDR category.
- Register the rule: Add your function to the PLURAL_RULES dictionary using its ISO 639-1 code as the key.
```py
# e.g polish
def polish_rule(count: int) -> str:
    if count == 1:
        return "one"
    if 2 <= count % 10 <= 4 and not (12 <= count % 100 <= 14):
        return "few"
    return "many"


# Registry mapping ISO codes to rule functions
PLURAL_RULES = {
    "pl": polish_rule,
    # add here new set of rules
}
```
- Provide dictionary based translations
In locale YAML (e.g `polish.yaml`) use dictionary for `months` and `years` fields (keys have to match strings returned rule function!)
```py
# e.g polish
  months: 
    one: "miesiąc"
    few: "miesiące"
    many: "miesięcy"
  year: "rok"
  years:
    one: "rok"
    few: "lata"
    many: "lat"
```

Done!
## Others
There is a `default_rule` as a fallback even if a specific rule of key is missing.

There is backward compatibility for simple string-based locales (English, German, etc.).


sinaatalay:
Thank you. I'll review this before v2.8 release.

Issue #673, opened 2026-02-18
set the file as writable for user when copying a template


teto:
Using rendercv 2.6 from my linux distribution nixos ( https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/re/rendercv/package.nix), I get the following error when creating a new theme:

```
$ rendercv create-theme moderncvmatt
...
│ /nix/store/2ph4j9j8pzvxw90qmfy927gy6pz8g6rz-rendercv-2.6/lib/python3.13/site-packages/rende │
│ rcv/cli/create_theme_command/create_init_file_for_theme.py:43 in create_init_file_for_theme │
│                                                                                             │
│   40 │   │   'theme: Literal["classic"] = "classic"',                                       │
│   41 │   │   f'theme: Literal["{theme_name}"] = "{theme_name}"',                            │
│   42 │   )                                                                                  │
│ ❱ 43 │   init_file_path.write_text(new_init_file_contents, encoding="utf-8")                │
│   44                                                                                        │

...

│                                                                                             │
│ ╭──────────────────────────────────────── locals ─────────────────────────────────────────╮ │
│ │     data = 'from typing import Literal\n\nimport pydantic\n\nfrom                       │ │
│ │            rendercv.schema.models.base im'+32157                                        │ │
│ │ encoding = 'utf-8'                                                                      │ │
│ │   errors = None                                                                         │ │
│ │  newline = None                                                                         │ │
│ │     self = PosixPath('/home/teto/cv/moderncvmatt/__init__.py')                          │ │
│ ╰─────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                         
│                                                                                             │
│ /nix/store/slhpx9glq7vl99bwi93bgrhn3syv98s1-python3-3.13.11/lib/python3.13/pathlib/_local.p │
│ y:537 in open                                                                               │
│                                                                                             │
│   534 │   │   """                                                                           │
│   535 │   │   if "b" not in mode:                                                           │
│   536 │   │   │   encoding = io.text_encoding(encoding)                                     │
│ ❱ 537 │   │   return io.open(self, mode, buffering, encoding, errors, newline)              │
│   538 │                                                                                     │
│   539 │   def read_text(self, encoding=None, errors=None, newline=None):                    │
│   540 │   │   """                                                                           │
│                                                                                             │
│ ╭──────────────────────────── locals ─────────────────────────────╮                         │
│ │ buffering = -1                                                  │                         │
[...]

PR #686, opened 2026-03-06
Fix new theme directory and files permissions

theobori:
Hello,

This PR contains changes to explicitly apply permissions to folders and files for theme creation. This is particularly to avoid issues such as #673 on immutable distributions.

I also formatted the modified file with ruff and black.

### DIFF
--- a/src/rendercv/cli/create_theme_command/create_theme_command.py
+++ b/src/rendercv/cli/create_theme_command/create_theme_command.py
@@ -1,3 +1,4 @@
+import os
 import pathlib
 import textwrap
 from typing import Annotated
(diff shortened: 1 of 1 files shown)


sinaatalay:
Thanks for the contribution and for reporting this!

While this PR was open, we were working through a batch of issues and ended up implementing a fix for this in commit 21e3cd7f. Our approach is slightly different: the fix lives in `copy_templates()` itself rather than in `create_theme_command.py`, so all callers benefit automatically. It also preserves existing permission bits by adding user-write (`|= S_IWUSR`) rather than overwriting with hardcoded values.

Since the fix has already landed, I'm closing this PR. Really appreciate you flagging the issue and putting together a fix. Thank you!

PR #696, opened 2026-03-14
Add TIME_SPAN placeholder functionality in entry templates

### LINKED ISSUE #547: [Question/Feature Request] What is the placeholder keyword for `time_span` in custom entry templates?
### Description

I am customizing my `cv.yaml` templates and have enabled the time span calculation feature for my experience section. However, I cannot find the correct placeholder keyword to display the calculated time span string within my custom `experience_entry` template.

### My Configuration

I have enabled time spans in my settings:

```yaml
design:
  show_time_spans_in:
    - experience
  time_span: "HOW_MANY_YEARS years HOW_MANY_MONTHS months"
```

### The Problem

I am trying to use this calculated string in my custom template. I have tried using `TIME_SPAN`, `DURATION`, `TOTAL_TIME` and `TIME_SPAN_STRING` as keywords, but none seems to work (the text either doesn't appear or is not replaced).

Here is the template I am using:

```yaml
    experience_entry:
      main_column: |-
        **COMPANY**, LOCATION #h(1fr) TIME_SPAN · DATE
        POSITION
        SUMMARY
        HIGHLIGHTS
      date_and_location_column: ""
```

### Question

What is the correct placeholder keyword (e.g., like `DATE`, `LOCATION`, `TITLE`) that I should use in the `main_column` string to inject the calculated time span?

If this variable is not currently exposed to the custom templates, is there a way to access it?

Thank you!

faridrashidi:
## Summary

- Expose `TIME_SPAN` as a first-class template placeholder so users can place the computed duration string (e.g., "2 years 3 months") independently in custom entry templates
- Add `START_DATE` and `END_DATE` as standalone placeholders for ranged entries
- When `TIME_SPAN` is explicitly used in a template, suppress the automatic time-span suffix on `DATE` to prevent duplication
- Update schema descriptions and docs to document the new placeholders

Closes #547.

## Example

Users can now write custom templates like:

```yaml
experience_entry:
  main_column: |-
    **COMPANY**, LOCATION #h(1fr) TIME_SPAN · START_DATE – END_DATE
    POSITION
    SUMMARY
    HIGHLIGHTS
  date_and_location_column: ""
```

## Test plan

- [x] `TIME_SPAN` populates correctly in custom experience entry template
- [x] `DATE` suppresses appended time span when `TIME_SPAN` is explicitly used
- [x] `TIME_SPAN` is removed cleanly when time spans are disabled
- [x] `START_DATE` and `END_DATE` format correctly in custom templates
- [x] Existing `DATE` behavior preserved when `TIME_SPAN` is not used
- [x] Schema regenerated and matches model descriptions
- [x] `just check` passes (ruff, ty, pre-commit)
- [x] Full test suite passes (1385 tests)


faridrashidi:
@sinaatalay, I took a look at the current CI failures on this PR, and the Windows jobs do not seem related to the subsection changes.

All 3 failing jobs:
- `Test with Py3.12 on windows`
- `Test with Py3.13 on windows`
- `Test with Py3.14 on windows`

are failing on the same upstream test:
- `tests/test_offline_whl.py::test_rendercv_renders_pdf_offline`

The failure happens during the wheel install step, before the actual render assertion runs.

From the current `main` version of that test, it builds the venv python path like this:

```python
venv_python = tmp_path / "venv" / "bin" / "python"
```

and then uses it here:

```python
run(["uv", "pip", "install", "--python", str(venv_python), f"{wheel}[full]"])
```

That path works on macOS/Linux, but on Windows the venv interpreter should be under `venv/Scripts/python.exe`, not `venv/bin/python`.

The Windows logs are consistent with that: the failing command is trying to install into a path ending in `venv\bin\python`, and `uv pip install --python ...` exits with status 2.

So I think the current Windows failures are coming from a cross-platform issue in `tests/test_offline_whl.py` on `main`, rather than from this work in this PR.

PR #697, opened 2026-03-15
Setup multi-platform docker build

quentinms:
See: https://docs.docker.com/build/ci/github-actions/multi-platform/

Avoids getting the following warning when running on an ARM mac, or having to pass `--arch amd64` when using Apple's [`container`](https://github.com/apple/container)

> WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested

Tested locally by running

```sh
docker build --platform linux/arm64,linux/amd64  -t render:qms -f Dockerfile .
```

And then verifying that I'm not getting the warning when

```sh
docker run -v "$PWD":/work -w /work render:qms render cv.yaml
```

It might need some tweaks when running in Github Actions, we'll have to test.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -124,14 +124,18 @@ jobs:
           images: |
             ghcr.io/${{ github.repository }}
 
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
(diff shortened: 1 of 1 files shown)


sinaatalay:
Thank you.

Issue #706, opened 2026-03-23
`bold_keywords` does not work if keyword starts or ends with a non-letter


Prometheus3375:
In 2.7 there was change that adds word boundary to the end and the start of each keyword.

https://github.com/rendercv/rendercv/blob/1ffb4c48903da419239773955d924447be05153e/src/rendercv/renderer/templater/string_processor.py#L41-L69

That usually works, but if bold keyword starts/ends with a non-letter, then word boundary expects previous/next char to be a letter.

For example, I have bold keyword `Tech stack:`. The new pattern will make it bold in `Tech stack:blabla`, but not in `Tech stack: blabla`.

I have a couple solutions to this.
1. Analyze both ends of each keyword. If starting char is a letter, prepend with `\b`. If it is a non-letter, do nothing. Similarly with the ending char.
2. Keep `bold_keywords` as-is, add a new setting `bold_patterns` that allows users to specify patterns themselves. This solution can be extended to `style_patterns` where user specifies a pattern and for each one a list of styles to apply: bold, italic, underlined, etc.

enfff:
This explains why I can't make make `C++` bold, thanks for sharing.

sinaatalay:
Thank you @enfff and @Prometheus3375 . The fix will be released in v2.9.

PR #709, opened 2026-03-25
Updated pattern matching algorithm, see #706

enfff:
Fix to issue [706](https://github.com/rendercv/rendercv/issues/706)

I tested it with the string`C++`, and the example provided by the user. In this case, both `Tech stack:` and `Tech stack: blabla` are getting bold. All tests in the repository seems to be passing, feel free to test it yourself.

image


sinaatalay:
Thank you.

PR #732, opened 2026-04-12
Feat: added ordinal day format for dates

AbdelRahmanRahal:
Previously in my CV's I would copy paste this line:
```yaml
single_date: DAYth MONTH_NAME, YEAR # * MAKE SURE TO UPDATE THE "th" IN "DAYth" TO THE CORRECT SUFFIX FOR THE DAY NUMBER
```

I am tired of doing that :p

I added DAY_ORDINAL to the date templates to support this.

It allows for something like this
```yaml
single_date: DAY_ORDINAL MONTH_NAME, YEAR
```

Which would show: 6th July, 2025.

It automatically adjusts for "1st", "2nd", and "3rd" correctly.

It's kind of hacky, but I did not want to import any extra dependencies for such a trivial feature. If there's a better way to natively achieve this that I do not know about, please let me know.


AbdelRahmanRahal:
Fixed the failing test case to include the new feature. Feel free to add a dedicated test case for it, or let me know and I'll add one.

That being said, the failing ty check is not on my part. A search tells me ty was updated and no longer needs those `# ty: ignore[missing-argument]` suppression comments. However, I don't wanna mess with anything without being sure. Let me know what to do next.

Issue #716, opened 2026-03-31
Add option to shorten date ranges within the same year (e.g. "Mar 2023 – Jun 2023" → "Mar – Jun 2023")


Ventus218:
### Summary

When a date range falls within the same year, it could be displayed more concisely by omitting the repeated year.

This could be implemented as an optional setting. In my opinion, it should be disabled by default to avoid unexpected changes to current behavior.

For example:

* Current: `Mar 2023 – Jun 2023`
* Proposed: `Mar – Jun 2023`

### Motivation

This would improve readability and save space, especially in dense CV layouts where multiple date ranges are shown.

### I can work on this!

If this feature would be useful, I’d be happy to work on an implementation—just let me know!

Ventus218:
I've looked a bit more into it and it is not that easy since we accept a template which could actually be anything.

I am thinking about an option that when flagged just removes YEAR from the template of the starting date, preserving the template of the end date.

Another possibility would be to let the user define a template to be used specifically in case of date ranges that fall within the same year.

PR #735, opened 2026-04-18
Add option to shorten date ranges within the same year

Ventus218:
I've described the proposed feature here #716

Let me know if there's something to change or fix!


Ventus218:
Sorry to bother you @sinaatalay. How am i supposed to deal with the fact that the rendercv-skill is stale?
That's the only test not passing. And running the suggested script updates the submodule to which i obviously cannot push.

Thank you for your time and help :)

PR #744, opened 2026-04-30
Use uv.lock as single source of truth for pre-commit tool versions

### LINKED ISSUE #743: `just check` fails due to ty version mismatch between uv.lock pin and pre-commit constraint
## Summary

Running `just check` fails because the two steps it executes resolve different versions of `ty`:

1. `uv run ty check` uses **ty 0.0.24** (exact version pinned in `uv.lock`) → **passes**
2. `pre-commit ty-check` uses **ty>=0.0.24** (declared in `.pre-commit-config.yaml`) → resolves to the latest available release → **fails**

The newer `ty` release has improved type inference and no longer needs the existing `# ty: ignore[...]` hints in several files. Those comments are now flagged as `error[unused-ignore-comment]`, which breaks the pre-commit step while the `uv` step keeps passing.

The result is that a clean checkout cannot pass `just check` out of the box, and the failure mode is non-obvious because only one of the two type-check invocations breaks.

## Steps to reproduce

1. Clone the repository on `main`.
2. Run `uv sync`.
3. Run `just check`.

## Expected behavior

Both type-check invocations use the same `ty` version, and `just check` either passes cleanly or fails consistently.

## Actual behavior

`uv run ty check` passes; `pre-commit ty-check` fails with `error[unused-ignore-comment]` on the locations listed below.

## Affected locations

- `src/rendercv/renderer/templater/entry_templates_from_input.py` — lines 201, 205, 206 (`# ty: ignore[invalid-argument-type]`)
- `tests/renderer/templater/test_date.py` — lines 687, 736 (`# ty: ignore[missing-argument]`)
- `tests/renderer/templater/test_string_processor.py` — line 147 (`# ty: ignore[missing-argument]`)

## Proposed fixes

Either of the following resolves the inconsistency:

- **Option A** — Remove the stale `# ty: ignore` comments. The newer `ty` no longer needs them, so the suppressions are dead weight regardless.
- **Option B** — Pin the pre-commit dependency to `ty==0.0.24` (instead of `ty>=0.0.24`) in `.pre-commit-config.yaml` to keep both environments in sync.

Option A is the more forward-looking fix; Option B preserves the current pin and defers the cleanup. A combination (pin + plan to bump and clean up later) is also reasonable.

## Context

Originally surfaced while investigating CI failures on #742.

ale-saglia:
Closes #743

## Problem

`.pre-commit-config.yaml` pinned tool versions independently from `uv.lock`, creating two separate sources of truth that could silently drift. The concrete failure reported in #743: the `ty-check` hook used `additional_dependencies ["ty>=0.0.24"]`, which pre-commit resolved to the latest available `ty` release — newer than the version locked in `uv.lock`.
The newer release tightened type inference, turning existing `# ty: ignore` comments into errors and making `just check` fail on a fresh clone.

The same structural problem affected `ruff`, which was pinned to a specific tag in the `astral-sh/ruff-pre-commit` external repo independently of `uv.lock`.

## Solution

Replace all external pre-commit repos for Python tools (`ruff`, `codespell`, `ty`) with local hooks that invoke tools via `uv run --frozen`. This way the version is always determined by `uv.lock` — no version number appears in `.pre-commit-config.yaml` itself.

- `ruff` and `ty` were already in the dev dependency group; their external repo blocks (`astral-sh/ruff-pre-commit`, and the `additional_dependencies` trick for `ty`) are replaced by `language: system` hooks.
- `codespell` is added to the dev dependency group and locked in `uv.lock`, then made a local hook as well.
- `pre-commit/pre-commit-hooks` remains an external repo since it has no equivalent package in the project.

Codespell configuration (skip patterns, exclude file) is moved from inline hook args into `[tool.codespell]` in `pyproject.toml`, so the settings apply consistently regardless of how codespell is invoked.
The skip list is also extended to cover PDF files (which produce binary false positives) and the bundled `typst_fontawesome` third-party library.

## Testing

`uv run --frozen codespell` exits clean.  
`uv run --frozen prek validate-config .pre-commit-config.yaml` reports
success.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,24 +4,28 @@ repos:
     hooks:
       - id: check-added-large-files
       - id: check-toml
-  - repo: https://github.com/codespell-project/codespell
-    rev: v2.4.2
+  - repo: local
     hooks:
       - id: codespell
+        name: codespell
+        language: system
+        entry: uv run --frozen codespell
         args:
           - --skip=src/rendercv/schema/models/locale/other_locales/*
           - --exclude-file=schema.json
-  - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.7
-    hooks:
       - id: ruff-check
+        name: ruff check
+        language: system
+        entry: uv run --frozen ruff check --force-exclude
+        types_or: [python, pyi]
       - id: ruff-format
-  - repo: local
-    hooks:
+        name: ruff format
+        language: system
+        entry: uv run --frozen ruff format --force-exclude
+        types_or: [python, pyi]
       - id: ty-check
-        name: ty-check
-        language: python
-        entry: ty check src tests
+        name: ty check
+        language: system
+        entry: uv run --frozen ty check src tests
         pass_filenames: false
         args: [--python=.venv/]
-        additional_dependencies: ["ty>=0.0.24"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,9 +10,6 @@ repos:
         name: codespell
         language: system
         entry: uv run --frozen codespell
-        args:
-          - --skip=src/rendercv/schema/models/locale/other_locales/*
-          - --exclude-file=schema.json
       - id: ruff-check
         name: ruff check
         language: system
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -226,4 +226,5 @@ addopts = [
 testpaths = ["tests"]
 
 [tool.codespell]
-skip = "*.md"
+skip = "*.md,*.pdf,*/typst_fontawesome/*,*/other_locales/*"
+exclude-file = "schema.json"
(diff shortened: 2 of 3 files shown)


aimansour:
Verified this PR against the current `main` (`1d4b87b`) on 2026-09-18.

Current behavior is still reproducible:

- `uv run --frozen --all-extras ty check src tests` uses the locked Ty 0.0.24 and passes.
- The isolated pre-commit hook resolves the open constraint `ty>=0.0.24` to Ty 0.0.82 and fails with 39 diagnostics (19 `unused-ignore-comment`, 12 `invalid-argument-type`, and 8 `invalid-type-form`).

I then merged this PR into the current `main` in a clean worktree with no conflicts, ran `uv sync --frozen --all-extras`, and ran the full `just check`. Ruff, the locked Ty check, codespell, formatting, and every pre-commit hook passed.

So the version-drift problem remains active, and this PR's single-source-of-truth approach is still applicable to the current codebase. It would also prevent the same unrelated pre-commit failure currently appearing on newer external PRs.

PR #756, opened 2026-07-03
Add Itch.io and Devpost as a social network

TendTo:
Note that, the way tests are setup, the [rendercv-skill repository](https://github.com/rendercv/rendercv-skill.git) should be updated to add the new social networks.


Saka-CS:
I would love for this to be added

PR #605, opened 2025-12-26
fix: allow PDF/PNG generation when dont_generate_typst is True

majiayu000:
## Summary
- When `dont_generate_typst: true` is set, PDF and PNG files are now generated using a temporary Typst file
- The temporary file is automatically cleaned up after compilation
- This decouples Typst file output from PDF/PNG generation capabilities

## Changes
- `generate_typst()` now returns `tuple[pathlib.Path, bool]` - the path and whether it's temporary
- `generate_pdf()` and `generate_png()` no longer check for None typst_path
- `run_rendercv()` handles cleanup of temporary directories in a finally block
- Updated tests to reflect the new expected behavior

## Test plan
- [x] All existing tests pass (887 passed)
- [x] CLI test updated: when `dont_generate_typst=True`, PDF and PNG are generated, only `.typ` file is missing
- [x] Temporary directory is properly cleaned up after generation

Fixes #550

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #760, opened 2026-07-12
Scope temp Typst cleanup to compile call, fix #730

### LINKED ISSUE #730: PDF file isn't generated if dont_generate_typst is set to true
Hi there!

If settings.render_command.dont_generate_typst is set to true and settings.render_command.dont_generate_pdf is set to false, PDF file won't be rendered.

```shell
$ rendercv -v
RenderCV v2.8
```

YAML sample configuration:
```yaml
settings:
  current_date: today
  render_command:
    output_folder: output
    typst_path: OUTPUT_FOLDER/NAME_IN_SNAKE_CASE_CV.typ
    pdf_path: OUTPUT_FOLDER/NAME_IN_SNAKE_CASE_CV.pdf
    markdown_path: OUTPUT_FOLDER/NAME_IN_SNAKE_CASE_CV.md
    html_path: OUTPUT_FOLDER/NAME_IN_SNAKE_CASE_CV.html
    png_path: OUTPUT_FOLDER/NAME_IN_SNAKE_CASE_CV.png
    dont_generate_markdown: false
    dont_generate_html: true
    dont_generate_typst: false
    dont_generate_pdf: false
    dont_generate_png: true
  pdf_title: NAME - CV
```

vishnujayvel:
## Problem

Closes #730 (reported by @AndDeluiz). When `settings.render_command.dont_generate_typst` is `True` and PDF/PNG generation is enabled, `generate_pdf`/`generate_png` silently skip compilation instead of producing output. The Typst source is required to compile the PDF/PNG, but the code path that should provide it when the standalone `.typ` file isn't being persisted was never wired up.

As @sinaatalay noted on the issue, the right fix is to generate the Typst source into a temporary location whenever it's needed for compilation, regardless of whether the user wants the standalone file kept around.

## Fix

`ensure_typst_source_for_compilation` is now a `@contextlib.contextmanager` backed by `tempfile.TemporaryDirectory()`. `generate_pdf`/`generate_png` use it with a `with` block, so the temporary Typst source directory is created, used for compilation, and removed within the scope of that single compile call — on both success and error — rather than relying on a longer-lived temp dir.

This directly addresses the concern @sinaatalay raised on #577 ("Are we sure those temp dirs get removed at the end of the run?"), which along with the related #605 was closed over uncertainty about whether `mkdtemp` + `atexit`-based cleanup actually ran. Scoping cleanup to a context manager means it happens deterministically at the end of the `with` block, not deferred to process exit.

Also corrected the `dont_generate_typst` field description and CLI help text, which described it as controlling whether Typst (and therefore PDF/PNG) generation happens at all — it only controls whether the standalone `.typ` file is persisted to disk. Regenerated `schema.json` via `just update-schema` to match.

## Changes

- `src/rendercv/renderer/pdf_png.py`: `ensure_typst_source_for_compilation` converted to a context manager over a `TemporaryDirectory`; `generate_pdf`/`generate_png` updated to use it via `with`.
- `src/rendercv/schema/models/settings/render_command.py` and `src/rendercv/cli/render_command/render_command.py`: corrected `dont_generate_typst` description/help text.
- `schema.json`: regenerated from the corrected model.
- `tests/renderer/test_pdf_png.py`: added coverage for temp dir cleanup on both success and error, confirmed a persisted `typst_path` is left untouched, and an end-to-end check that `generate_pdf` leaves no temp dir behind.

## Verification

- `uv sync --frozen --all-extras` — installs cleanly (submodules, including `typst_fontawesome`, need `git submodule update --init --recursive` first).
- `uv run --frozen --all-extras pytest tests/renderer/test_pdf_png.py tests/cli/render_command/test_render_command.py` — **56 passed**, 0 failed.
- `uv run --frozen --all-extras ruff check` on all files touched by this patch — **all checks passed**.


vishnujayvel:
The failing `Run pre-commit checks` job looks unrelated to this PR: the prek hook installs floating `ty>=0.0.24` (currently resolving to 0.0.61) while `uv.lock` pins 0.0.24, and the newer ty's 36 diagnostics (`unused-ignore-comment`, stricter `invalid-type-form`/`invalid-argument-type`) reproduce identically on unmodified `main` — none of them are in files this PR touches. Happy to open a separate, scoped PR upgrading ty and clearing those diagnostics if that's useful; keeping it out of this one so the diff stays focused on #730. (The `review` job failure is org-side — empty `ANTHROPIC_API_KEY` in the workflow.)

PR #764, opened 2026-07-23
Add plain text output format

### LINKED ISSUE #763: Plain text (.txt) output format
## Problem

RenderCV currently generates PDF, Typst, Markdown, HTML and PNG outputs. There is no plain text output.

Markdown often plays the role of the "text version" of a CV, but it only reads well when rendered, or when opened in an editor with Markdown support. Read as raw text, its markup gets in the way: `| **Label** | value |` table rows, `[text](url)` links, `**bold**` markers and so on. For anyone opening the file in a pager, a basic editor, a terminal, an email client or an ATS pipeline, raw Markdown is not very readable.

Plain text does not have this problem. Tables, section separators and more complex layouts can be expressed with spaces and simple ASCII rules, and the result reads the same everywhere, with no specific editor or renderer required.

## Proposal

Add a `text` output format alongside the existing ones:

- a `generate_text()` renderer following the same architecture as the Markdown one (Jinja templates under `templater/templates/text/`, rendered from the same processed model)
- `text_path` and `dont_generate_text` in `settings.render_command`
- `--text-path` and `--dont-generate-text` CLI options
- links rendered as `label (url)` when the URL adds information over the link text, emphasis and code markers stripped, section titles underlined
- user template overrides working exactly as for the other formats, so the txt output can be restyled (80-column layouts, ASCII tables, custom headers) without touching the core

I have a working implementation with tests, regenerated schema and examples, and updated docs. I will open a PR shortly and link it here.

eldios:
Closes #763

## What

Adds a plain text (`.txt`) output format alongside the existing PDF, Typst, Markdown, HTML and PNG outputs, as proposed in #763.

- New `generate_text()` renderer (`renderer/text.py`), same structure as the Markdown renderer
- New Jinja template set under `templater/templates/text/` (header, section begin/end, one template per entry type)
- `markdown_to_plain_text()` in the Markdown parser: emphasis and code markers stripped, links rendered as `label (url)` when the URL adds information over the link text
- Plain text connection formatting (`compute_connections_for_text`)
- `text_path` and `dont_generate_text` settings, `--text-path` / `--dont-generate-text` CLI options, wired through the model builder
- Entry templates now receive `entry_index` and `entries_count`, so numbered entries can carry real numbers in formats that have no renderer doing the numbering; existing templates do not use them, so there is no behavior change
- Docs updated (user guide index and settings page), JSON schema and example YAMLs regenerated

The regenerated binary artifacts (example PDFs, PNGs, Typst files) are intentionally left out to keep the diff reviewable; the generated-files tests only require the example YAMLs and schema.json to be current.

User template overrides work exactly as for the other formats: placing a `text/` directory next to the input file allows full restyling of the txt output (80-column layouts, ASCII tables, custom headers) without touching the core.

## Test plan

Ran the full test suite and regenerated the reference files with `just update-testdata`, which produced the new `tests/renderer/testdata/test_text/` references for the minimal and full models. Added unit tests for `markdown_to_plain_text()`. Also rendered a complete real-world CV to text and reviewed the output side by side with the Markdown and PDF versions.


eldios:
About the failing pre-commit check: the ty-check hook installs ty with >=0.0.24, so it picks up the latest release, and recent ty versions report errors on files this PR does not touch. Running the same hook locally with ty pinned to the version resolved in uv.lock (0.0.24), the whole pre-commit suite passes, ty-check included. I left the hook configuration untouched here since it is not part of this change; happy to update the pin in a separate PR if useful.

PR #774, opened 2026-08-25
Add appending to list in overrides

NoelJacob:
In ovverrides you cannot append to a list. This PR makes it possible. Tests also fixed. Minimal changes.


NoelJacob:
The failing tasks are preexisting as it is not regarding the code I modified.

PR #781, opened 2026-09-09
Fix RTL phone number rendering in PDFs

NullUser3:
## Description

Fix Arabic phone number rendering in the Typst renderer.

When using the Arabic locale, phone numbers containing Latin digits and the `+` prefix were rendered in the wrong direction due to RTL text handling. This change ensures that phone numbers are rendered correctly when the CV is using Arabic.

## What changed

- Fixed the RTL/LTR handling of phone numbers in the Typst renderer.
- Ensured the `+` prefix and phone number digits appear in the correct order.
- Added/updated the relevant test to prevent regressions.

## Before

The phone number is displayed in the incorrect order when rendered in Arabic.

before

> The CV uses fictional data for demonstration purposes.

## After

The phone number is displayed in the correct order.

after

> The CV uses fictional data for demonstration purposes.

## Testing

- Ran the test suite with:
  `just test`
- Verified the generated Arabic CV PDF manually.


NullUser3:
Note: The two remaining CI errors are pre-existing. I verified that they also occur on the base branch before applying the changes in this PR, so they are unrelated to this contribution.

PR #785, opened 2026-09-17
Fix AI agent skill theme coverage

### LINKED ISSUE #784: AI agent skill advertises only 6 of 9 built-in themes
### Problem

RenderCV 2.8 accepts nine built-in themes, but the generated AI agent skill advertises only six. The missing themes are ember, ink, and opal.

The generator intentionally limits detailed theme overrides to keep the skill context compact. However, the same hard-coded subset is also passed to the template as the complete available-themes list, so the skill incorrectly tells agents that only six themes exist.

### Root cause

scripts/rendercv_skill/generate.py uses SKILL_THEMES both for:

1. the public list of available themes; and
2. the subset whose detailed override YAML is embedded.

These are separate concerns. The engine already exposes the authoritative dynamic list through available_themes.

### Proposed fix

- Use rendercv.schema.models.design.built_in_design.available_themes for the skill metadata and Quick Start list.
- Rename the compact hard-coded subset to THEMES_WITH_DETAILED_OVERRIDES and use it only for expanded override examples.
- Clarify in the template that detailed examples cover a representative subset.
- Add a regression test comparing the themes advertised by the generated skill with the themes accepted by RenderCV.

I have implemented and tested this locally. The generated skill grows by only about 108 bytes because the three themes are listed without embedding their detailed override YAML.

### Contribution workflow question

The generated SKILL.md lives in the read-only rendercv/rendercv-skill submodule, and the main repository's generated-file test requires that submodule content to be updated. External contributors do not have push access to the distribution repository, and it explicitly asks contributors to work in rendercv/rendercv.

What is the preferred workflow for a PR that changes the generated skill? I can open the main-repository PR immediately once there is a maintainer-supported way to publish the matching submodule commit.

aimansour:
Closes #784.

## Summary

- derive the skill's advertised theme list from the engine's `available_themes` instead of a hard-coded subset
- keep detailed override examples limited to a representative subset so the skill remains compact
- add a regression test ensuring every engine-supported theme is advertised
- regenerate the checked-in skill documentation and downloadable ZIP

This adds the previously omitted `ember`, `ink`, and `opal` themes to the public list while preserving the existing context-size optimization for detailed examples.

## Verification

- `uv run pytest tests/test_generated_files.py -q` — 11 passed, 1 skipped (Windows-specific schema-generation skip)
- `uv run ruff check scripts/rendercv_skill/generate.py tests/test_generated_files.py`
- `uv run ruff format --check scripts/rendercv_skill/generate.py tests/test_generated_files.py`
- `uv run ty check tests/test_generated_files.py`
- full suite with the unrelated pre-existing Hypothesis failure `TestRemoveNotProvidedPlaceholders::test_missing_placeholders_removed` deselected — 1536 passed, 2 skipped

`mkdocs build --clean --strict` currently hits an unrelated baseline issue after initializing the `typst_fontawesome` submodule: the API-reference generator discovers `typst_fontawesome/helper.py`, but Griffe cannot collect that non-package submodule path.

## Maintainer action needed

The generated `.claude/skills/rendercv-skill/skills/rendercv/SKILL.md` lives in the read-only `rendercv/rendercv-skill` distribution repository. External contributors cannot publish its new commit, so this PR intentionally does not change that submodule pointer. Please publish the regenerated `SKILL.md` there (its content matches the updated `docs/llms.txt`) and then update the submodule pointer in this PR. Maintainer edits are welcome.


aimansour:
Distribution update: the generated skill commit is now safely published on the fork at https://github.com/aimansour/rendercv-skill/commit/f5e035c. A maintainer can cherry-pick it into rendercv/rendercv-skill and then update this PR's submodule pointer.

PR #788, opened 2026-09-18
Require country clarification for local phone numbers

### LINKED ISSUE #787: AI skill may silently omit user-provided local phone numbers
### Problem

The AI agent skill currently says:

> If the user provides a local number without country code, ask which country, or omit the phone field.

RenderCV requires international phone numbers, while `cv.phone` is optional. This allows an agent to silently discard a phone number the user explicitly provided instead of resolving the missing country code.

### Expected behavior

For a local number, the agent should ask which country it belongs to. It should neither guess the country code nor omit the number without the user's explicit permission.

### Example

A request such as `Create a CV for Sam Lee. Their phone number is 020 7946 0958.` should trigger a country clarification before the final CV is generated.

### Proposed fix

Make the instruction deterministic and add an agent eval covering the clarification, no-guessing, and no-silent-omission requirements.

aimansour:
## Summary

- require country clarification for local phone numbers
- prohibit guessing a country code or silently omitting a user-provided number
- add an agent eval for the clarification behavior
- regenerate `docs/llms.txt` and the downloadable skill archive

Fixes #787.

## Verification

- `uv run pytest tests/test_generated_files.py -q` (10 passed, 1 skipped on Windows)
- parsed `scripts/rendercv_skill/evals/tests/cv_generation.yaml` with ruamel.yaml

The new promptfoo case was not executed locally because no Anthropic API credential is available. A companion update to the generated `rendercv-skill` distribution repository will be linked separately.


aimansour:
Distribution note: the matching generated `rendercv-skill` commit is available at https://github.com/aimansour/rendercv-skill/commit/e6e63ab597679865485fc4abdc27a0e68acd9c8f. GitHub rejected opening a pull request against `rendercv/rendercv-skill` for this account (GraphQL permission error / REST 404), so a maintainer will need to cherry-pick that commit or run the generator and update the submodule pointer before the generated-file check can pass.

PR #790, opened 2026-09-18
Fix serialization of multiple phone numbers

### LINKED ISSUE #789: Serializing multiple phone numbers raises AttributeError
### Problem

`cv.phone` accepts either one phone number or a list, but serializing a validated list crashes because `Cv.serialize_phone` calls `.replace()` on the list itself.

### Reproduction

```python
from rendercv.schema.models.cv.cv import Cv

cv = Cv.model_validate({
    "name": "Test",
    "phone": ["+14155552671", "+442071838750"],
})
cv.model_dump()
```

Current result:

```text
pydantic_core.PydanticSerializationError: Error calling function `serialize_phone`:
AttributeError: 'list' object has no attribute 'replace'
```

This also affects the AI skill because its generated schema advertises multiple phone numbers as supported.

### Expected behavior

Serialization should remove the `tel:` prefix from each validated phone number while preserving scalar and `None` behavior.

### Suggested fix

Handle scalar and list values separately in the serializer and add regression coverage for both shapes.

aimansour:
## Summary
- serialize every validated phone number when `cv.phone` is a list
- preserve the existing scalar and `None` behavior
- add a regression test covering two international phone numbers

Fixes #789.

## Verification
- `uv run pytest tests/schema -q` — 896 passed
- `uv run pytest -q -k "not test_missing_placeholders_removed"` — 1536 passed, 2 skipped
- `uv run ruff check .` — passed
- `uv run ty check src/rendercv/schema/models/cv/cv.py tests/schema/models/cv/test_cv.py` — passed

The unfiltered suite has one unrelated existing Hypothesis failure: `test_missing_placeholders_removed` can generate `EF`, then incorrectly asserts that `EF` is absent from the literal text `PREFIX SUFFIX`.


aimansour:
CI follow-up: all nine pytest matrix jobs (Ubuntu, macOS, and Windows on Python 3.12–3.14) passed, and coverage passed. The two red checks are repository-wide workflow issues unrelated to this change: the unpinned pre-commit Ty environment reports 39 diagnostics in existing files, and the Claude workflow cannot obtain an OIDC token because `id-token: write` is missing.

